PR #12, opened 2025-08-20
Add Crossplane Azure Blob Storage claim template and associated tests

jaraya-mycarrier:


jaraya-mycarrier:
Improvements based on Copilot review:
  - Extracted duplicated storage account name generation logic into reusable helper templates
  - Extracted duplicated resource group name generation logic into helper template
  - Simplified complex nested default expressions for better readability
  - Moved Crossplane-specific helpers to dedicated _helpers.crossplane.tpl file
  - Removed unused commented code (writeConnectionSecretsToRef section)

PR #88, opened 2026-05-29
fix(env-merge): only strip KeyVault_* env keys for csharp apps

pkuzmenko-mycarrier:
## Summary
- Fixes a 3.1.12 regression where the new `helm.application.env` merge helper deleted `KeyVault_*` env vars for **non-csharp** apps.
- The helper applied its omit list to both `global.env` and `application.env`. The 5 `KeyVault_*` keys are only re-injected by `helm.lang.vars.csharp` (emits only when `language == csharp`). For nodejs apps nothing re-injected them, so keys sourced from `global.env` were stripped with no replacement.
- **Impact:** MC.MyCarrier `feature22`/`feature23` deployments stuck on missing `KeyVault_SplitIoProxyApiKey`; revert to 3.1.11 resolved it.
- **Fix:** gate the `KeyVault_*` omits on `language == csharp`; keep always-managed computed keys (`OTEL_EXPORTER_OTLP_ENDPOINT`, `ComputedEnvironmentName`, `ActiveOffloads`) omitted unconditionally. Chart bumped `3.1.12` → `3.1.13`.

## Verification
Rendered MC.MyCarrier `feature22` merged values against buggy `3.1.12` vs fixed `3.1.13`:

| | buggy 3.1.12 | fixed 3.1.13 |
|---|---|---|
| container env var count | 139 | 143 |
| `KeyVault_SplitIoProxyApiKey` | missing | present |
| `KeyVault_SplitIoProxyUrl` / `KeyVault_IsActive` / `KeyVault_RedisConnection` | missing | present |
| vars lost vs buggy | — | none |

Fix restores exactly the +4 keys, nothing else changes.

## Test plan
- [x] `helm unittest charts/mycarrier-helm` — 504/504 pass (54 suites)
- [x] 3 new regression tests in `environment_variables_test.yaml`:
  - nodejs keeps `global.env.KeyVault_SplitIoProxyApiKey`
  - csharp strips the user copy (single source = lang.vars), no duplicate
  - nodejs still omits computed `ComputedEnvironmentName` from `global.env`
- [x] Local render diff of MC.MyCarrier feature22 (buggy vs fixed)


charts/mycarrier-helm/templates/_environment.tpl
```diff
@@ -121,14 +121,21 @@ They use a straightforward default route without header matching requirements.
...
 where the call-site dot exposes both `.Values.global.env` and `.application.env`.
 */}}
 {{- define "helm.application.env" -}}
-{{- $omitKeys := list "OTEL_EXPORTER_OTLP_ENDPOINT" "ComputedEnvironmentName" "ActiveOffloads" "KeyVault_RedisConnection" "Auth_KeyVault_RedisConnection" "KeyVault_IsActive" "KeyVault_SplitIoProxyApiKey" "KeyVault_SplitIoProxyUrl" -}}
+{{- $omitKeys := list "OTEL_EXPORTER_OTLP_ENDPOINT" "ComputedEnvironmentName" "ActiveOffloads" -}}
+{{- if eq (lower (toString $.Values.global.language)) "csharp" -}}
+{{- $omitKeys = concat $omitKeys (list "KeyVault_RedisConnection" "Auth_KeyVault_RedisConnection" "KeyVault_IsActive" "KeyVault_SplitIoProxyApiKey" "KeyVault_SplitIoProxyUrl") -}}
+{{- end -}}
 {{- $merged := dict -}}
 {{- with $.Values.global.env -}}
```
hpyvovar-mycarrier: The omit gate strips `KeyVault_RedisConnection` and `Auth_KeyVault_RedisConnection` unconditionally for all csharp apps, but `helm.lang.vars.csharp` only re-injects them when `global.dependencies.redis: true` (`_lang.tpl:74-81`). Since `dependencies.redis` defaults to `false`, a csharp app that sets these keys via `global.env` without enabling the redis dependency will silently lose them — the same class of regression this PR fixes for nodejs, left open for csharp.

Gate the redis pair on the same condition that drives their re-injection:

```suggestion
{{- $omitKeys := list "OTEL_EXPORTER_OTLP_ENDPOINT" "ComputedEnvironmentName" "ActiveOffloads" -}}
{{- if eq (lower (toString $.Values.global.language)) "csharp" -}}
{{- $omitKeys = concat $omitKeys (list "KeyVault_IsActive" "KeyVault_SplitIoProxyApiKey" "KeyVault_SplitIoProxyUrl") -}}
{{- if dig "dependencies" "redis" false $.Values.global -}}
{{- $omitKeys = concat $omitKeys (list "KeyVault_RedisConnection" "Auth_KeyVault_RedisConnection") -}}
{{- end -}}
{{- end -}}
```

hpyvovar-mycarrier (changes requested):

hpyvovar-mycarrier (approved):

charts/mycarrier-helm/templates/_environment.tpl
pkuzmenko-mycarrier: Good catch — confirmed real. `helm.lang.vars.csharp` only re-injects `KeyVault_RedisConnection` / `Auth_KeyVault_RedisConnection` under `{{ if .Values.global.dependencies.redis }}` (`_lang.tpl:74-81`), and `dependencies.redis` defaults to false, so stripping them unconditionally for csharp would silently drop a `global.env`-supplied value when redis is off — same regression class.

Fixed in `cfaab7a`: the redis pair is now gated on `dig "dependencies" "redis" false $.Values.global`, matching where it's re-injected. The other three csharp keys (`KeyVault_IsActive`, `KeyVault_SplitIoProxyApiKey`, `KeyVault_SplitIoProxyUrl`) are injected unconditionally in the csharp block, so they remain unconditionally omitted. Added two regression tests (csharp + redis-off keeps the user value; csharp + redis-on strips the user copy so lang.vars is the single source). `helm unittest`: 506/506 pass.

PR #101, opened 2026-07-27
feat(mycarrier-helm): v3.5.0 secure defaults — drop-caps + readiness/liveness probes on (DEVOPS-176 Workstream B)

pkuzmenko-mycarrier:
## Summary

Three secure-by-default, per-app-overridable changes to `charts/mycarrier-helm`, bumped to `v3.5.0`:

1. **Drop all Linux capabilities by default** — `_podsecurity.tpl` (`helm.containerSecurityContext`) now adds `capabilities.drop: [ALL]` to every container securityContext. Services that need a specific capability back can opt in via `securityContext.addCapabilities: [...]` in their application values (satisfies `disallow-add-capabilities` for services that don't opt in).
2. **Deployment readiness probe on by default** — `_spec_deployment.tpl`: `probes.enableReadiness` now defaults `true` (was `false`), matching the existing default-true pattern already used for liveness/startup. Still overridable per app via `probes.enableReadiness: false`.
3. **StatefulSet + Rollout probes on by default** — `_spec_statefulset.tpl` / `_spec_rollout.tpl`: liveness/readiness/startup probes were previously fully opt-in (required explicit `enableX: true`); now default to `true` via the same `dig "probes" "enableX" true .application` pattern the Deployment template already used. All three remain overridable per app (`probes.enableLiveness/enableReadiness/enableStartup: false`).

All three changes are **per-application overridable** via `values.yaml` — no service is forced into new behavior it can't opt out of.

Publishing chart `3.5.0` does **not** deploy anywhere by itself — nothing consumes it until the `workflow-dev` chart-version pin is bumped, which is a separate, explicitly gated rollout step (not part of this PR; `workflow-dev`/`workflow-core` were not touched here).

Seccomp profile enforcement is intentionally deferred to a follow-up — out of scope for this change.

## Validation

- `helm lint charts/mycarrier-helm` — pass.
- `helm unittest charts/mycarrier-helm` (existing suite) — 570/570 tests pass. This caught and led to a fix for a regression where `.application` is `nil` in the Job/CronJob template context (capabilities dig now guarded with `{{- if .application }}`, matching the existing guard pattern used for `readOnlyRootFilesystem`).
- **Render-diff**: rendered the chart before/after these changes using MC.Address-derived values (all 9 real Deployment apps) plus synthetic StatefulSet/Rollout/override-capability test cases. The diff shows *only*:
  - added `readinessProbe` (tcpSocket) on Deployment containers that didn't already have one (11 additions: 9 Deployment apps + StatefulSet + Rollout example)
  - added `livenessProbe`/`startupProbe` on the StatefulSet/Rollout examples (2 each)
  - added `capabilities: {drop: [ALL]}` on every container (12 additions)
  - added `capabilities.add: [NET_BIND_SERVICE]` on the one synthetic app that set `securityContext.addCapabilities` (confirming override works)
  - `targetRevision: 3.4.0 → 3.5.0` in the embedded ArgoCD Application example (expected consequence of the Chart.yaml bump)
  - No other fields changed, reordered, or broken.

## Test plan

- [x] `helm lint charts/mycarrier-helm` passes
- [x] `helm unittest charts/mycarrier-helm` — 570/570 pass
- [x] Render-diff (old 3.4.0 vs new 3.5.0 templates) reviewed — only the three intended changes present
- [ ] Reviewer sign-off
- [ ] Separate, explicitly-gated `workflow-dev` chart-version pin bump before this reaches any live deployment

### DIFF
--- a/charts/mycarrier-helm/templates/_podsecurity.tpl
+++ b/charts/mycarrier-helm/templates/_podsecurity.tpl
@@ -16,6 +16,15 @@ securityContext:
   runAsNonRoot: true
   readOnlyRootFilesystem: {{ if and .application .application.securityContext .application.securityContext.readOnlyRootFilesystem }}true{{ else }}false{{ end }}
   allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+      - ALL
+{{- if .application }}
+{{- with (dig "securityContext" "addCapabilities" (list) .application) }}
+    add:
+{{ toYaml . | indent 6 }}
+{{- end }}
+{{- end }}
 {{- end }}
 {{- end -}}
 
--- a/charts/mycarrier-helm/Chart.yaml
+++ b/charts/mycarrier-helm/Chart.yaml
@@ -3,5 +3,5 @@ name: mycarrier-helm
 description: A Helm chart for MyCarrier GitOps
 type: application
 icon: https://go.mycarrier.io/hs-fs/hubfs/Logos/PNG-MC-COLOR-Logo.png?width=366&height=85&name=PNG-MC-COLOR-Logo.png
-version: 3.4.0
+version: 3.5.0
 appVersion: 3.0.29
--- a/charts/mycarrier-helm/templates/_spec_deployment.tpl
+++ b/charts/mycarrier-helm/templates/_spec_deployment.tpl
@@ -122,7 +122,7 @@ template:
         {{ include "helm.defaultLivenessProbe" . | indent 8 | trim }}
         {{- end }}
         {{- end }}
-        {{- if dig "probes" "enableReadiness" false .application }}
+        {{- if dig "probes" "enableReadiness" true .application }}
         {{- if and (dig "probes" false .application) (dig "readinessProbe" false .application.probes) }}
         readinessProbe:
           {{ toYaml .application.probes.readinessProbe | indent 10 | trim }}
--- a/charts/mycarrier-helm/templates/_spec_rollout.tpl
+++ b/charts/mycarrier-helm/templates/_spec_rollout.tpl
@@ -124,24 +124,24 @@ template:
             protocol: TCP
           {{- end }}
         {{- end }}
-        {{- if and .application.probes (hasKey .application.probes "enableLiveness") (.application.probes.enableLiveness) }}
-        {{- if .application.probes.livenessProbe }}
+        {{- if dig "probes" "enableLiveness" true .application }}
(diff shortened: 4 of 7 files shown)


charts/mycarrier-helm/templates/_podsecurity.tpl
```diff
@@ -16,6 +16,15 @@ securityContext:
   runAsNonRoot: true
   readOnlyRootFilesystem: {{ if and .application .application.securityContext .application.securityContext.readOnlyRootFilesystem }}true{{ else }}false{{ end }}
   allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+      - ALL
+{{- if .application }}
+{{- with (dig "securityContext" "addCapabilities" (list) .application) }}
+    add:
```
hpyvovar-mycarrier, line 24: **Test coverage gap**: no unit test exercises `securityContext.addCapabilities` / the new `capabilities.drop: [ALL]` rendering path. `tests/security_context_test.yaml` only covers `podSecurityContext`/`containerSecurityContext`, not this addition.

This gap is why an earlier version of this same block (with `| trim` after `indent 6`) rendered invalid YAML whenever `addCapabilities` was set — it slipped past `helm lint` and all 570 existing unittest cases since none of them render that path. That specific bug is already fixed in the current HEAD (`f0b96d5`, `trim` removed) — confirmed locally that it now renders valid YAML for single- and multi-item lists.

Recommend adding a case to `security_context_test.yaml` that sets `securityContext.addCapabilities` and asserts `spec.template.spec.containers[0].securityContext.capabilities.add`/`.drop`, so this path stays protected going forward.

pkuzmenko-mycarrier: Good catch — this path had zero coverage. Added three cases to `security_context_test.yaml`: (1) no `addCapabilities` set → asserts `capabilities.drop: [ALL]` and that `capabilities.add` doesn't render; (2) single-item `addCapabilities` → asserts `add`; (3) multi-item `addCapabilities` (`NET_BIND_SERVICE` + `SYS_TIME`) → asserts the multi-item `add` list, which is exactly the shape that broke under the earlier `| trim` variant you mention. Confirmed the fixed template (`_podsecurity.tpl`, no `trim`) renders valid YAML for all three via both `helm-unittest` and a manual `helm template`. Suite is now 573/573 (was 570), `helm lint` still clean. Test-only diff — no template change needed since HEAD was already correct. Thanks for flagging it.

hpyvovar-mycarrier (approved):

PR #105, opened 2026-07-31
fix(splitproxy): assert runAsNonRoot:true for kyverno enforce (DEVOPS-176)

pkuzmenko-mycarrier:
Part of DEVOPS-176. Adds pod-level `podSecurityContext.runAsNonRoot: true` to the splitproxy chart values (`charts/splitproxy/values.yaml`) + bumps `charts/splitproxy/Chart.yaml` 0.3.11→0.3.12. splitproxy already runs uid 1000 (no root override in any env); additive. Template already wires `.Values.podSecurityContext` to pod securityContext.

`helm lint` + `helm template` PASS (renders runAsNonRoot:true, chart label splitproxy-0.3.12).

**On merge, chart-releaser publishes 0.3.12 to charts.mycarrier.dev.** Paired with the ManagementInfra repin PR which MUST merge only AFTER this publishes.

### DIFF
--- a/charts/splitproxy/Chart.yaml
+++ b/charts/splitproxy/Chart.yaml
@@ -3,4 +3,4 @@ name: splitproxy
 description: A Helm chart for splitproxy server and admin-dashboard
 type: application
 icon: https://www.split.io/wp-content/uploads/2017/11/split-logo-light-background-transparent.png
-version: 0.3.11
+version: 0.3.12
--- a/charts/splitproxy/values.yaml
+++ b/charts/splitproxy/values.yaml
@@ -26,7 +26,8 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
-podSecurityContext: {}
+podSecurityContext:
+  runAsNonRoot: true
 
 securityContext: {}
 


bcarlock-mycarrier (approved):

charts/splitproxy/values.yaml
```diff
@@ -26,7 +26,8 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
-podSecurityContext: {}
+podSecurityContext:
+  runAsNonRoot: true
```
hpyvovar-mycarrier, line 30: **🔴 Critical — `runAsNonRoot: true` without `runAsUser` guarantees `CreateContainerConfigError` on every splitproxy pod.**

The deployed image `splitsoftware/split-proxy:5.5.1` declares a **non-numeric** user in its image config — verified against the Docker Hub registry (config digest `sha256:20408dd8…`: `"User": "split-proxy"`; build history: `adduser --system --uid 1000 'split-proxy'` then `USER split-proxy`).

Kubelet's `verifyRunAsNonRoot` cannot resolve named users: with effective `runAsUser` unset and a non-numeric image USER, the container fails with:

```
CreateContainerConfigError: container has runAsNonRoot and image has non-numeric user (split-proxy), cannot verify user is non-root
```

Kubelet never reads `/etc/passwd` inside the image — the PR body's "splitproxy already runs uid 1000" is true at runtime but irrelevant to this admission-time check, which only accepts a numeric UID from the image config or an explicit `runAsUser`. The chart wires `podSecurityContext` verbatim into `spec.template.spec.securityContext` (templates/deployment.yaml), no downstream consumer sets `runAsUser`, and kyverno-policies has no mutate rule to inject one.

Once 0.3.12 syncs via ManagementInfra#147, all generated splitproxy Applications (`development-{dev,qa,uat}` + `production-csp-prod`) wedge: new ReplicaSet pods sit in `CreateContainerConfigError`, old pods keep serving under the default rollout limits (k8s default `maxUnavailable: 25%`, not chart-configured) until any reschedule, drain, or HPA event erodes capacity. Permanent until reverted.

Same failure class: kube-logging/logging-operator#830, ribbybibby/ssl_exporter#102.

The fix is behaviourally identical to today's runtime state (the image resolves `split-proxy` → uid 1000 / gid 1000; `addgroup -g 1000` confirms the GID) and also gives the bank-vaults webhook-injected `copy-vault-env` init container a verifiable numeric identity:

```suggestion
podSecurityContext:
  runAsNonRoot: true
  runAsUser: 1000
  runAsGroup: 1000
```

charts/splitproxy/Chart.yaml
```diff
@@ -3,4 +3,4 @@ name: splitproxy
 description: A Helm chart for splitproxy server and admin-dashboard
 type: application
 icon: https://www.split.io/wp-content/uploads/2017/11/split-logo-light-background-transparent.png
-version: 0.3.11
+version: 0.3.12
```
hpyvovar-mycarrier, line 6: **💡 Suggestion — this chart bump changes pod admission semantics with zero rendered-output assertions; the green CI signal asserts nothing.**

This repo has helm-unittest wired up (`.github/workflows/unit-tests.yaml`), and three charts already carry suites (`mycarrier-helm`, `mc-environment`, `mongobetween`). `charts/splitproxy` has no `tests/` directory, so the workflow's no-suites fallback runs `helm unittest charts/splitproxy` and reports green — both `unittests (charts/splitproxy)` and `Helm Unit Test Results - charts/splitproxy` came back `success` on this head SHA having asserted nothing about the change.

Since the critical defect on this PR is a rendered-manifest problem, a three-assertion `charts/splitproxy/tests/deployment_test.yaml` would have caught it pre-merge:
- assert `spec.template.spec.securityContext.runAsNonRoot` is `true`
- assert `.runAsUser` is `1000`
- assert the image reference resolves to a non-empty tag

Recommend adding it alongside the `runAsUser` fix.

hpyvovar-mycarrier (changes requested):

PR #106, opened 2026-07-31
feat(mc-environment): forward cronjobs to mycarrier-helm (0.2.10)

pkuzmenko-mycarrier:
## What
- ApplicationSet generator (`appset.yaml`) now forwards `$env.cronjobs` to `mycarrier-helm`, mirroring the existing `jobs` block.
- `values.schema.json` adds an optional `cronjobs` property in 2 places: the shared `properties.cronjobs` `$ref` and the per-environment `oneOf` (`$ref: #/definitions/cronjobs` | `null`), both pointing at `mycarrier-helm`'s `values.schema.json#/properties/cronjobs`.
- Chart version bumped 0.2.9 -> 0.2.10.

## Why
`mc-environment` did not forward `cronjobs` to `mycarrier-helm`, blocking v2 CronJob rendering (needed for frontend looptesting). This fix is reusable across all `mc-environment` repos.

## Validation
- `helm lint` passes.
- Real `helm template` run on the patched chart accepts and forwards `cronjobs` (schema validates OK).
- `mycarrier-helm` renders `kind: CronJob` from the forwarded shape.
- Robustness review clean: `$ref` resolves, change is backward-compatible / no-op for repos without `cronjobs` defined, schema property is optional.

## MERGE ORDER (IMPORTANT)
This is step 1 of a hard-ordered set:
1. **Merge + publish 0.2.10 FIRST** (this PR).
2. Then the two consumer-pin PRs (workflow-dev + workflow-core, bump 0.2.9 -> 0.2.10).
3. Then the `MC.MyCarrier.Frontend` values PR.

Shipping the frontend `cronjobs` values while consumers still pin 0.2.9 hard-fails Helm schema validation.

Refs DEVOPS-176.

### DIFF
--- a/charts/mc-environment/templates/appset.yaml
+++ b/charts/mc-environment/templates/appset.yaml
@@ -36,6 +36,9 @@ spec:
                 {{ with $env.jobs }}jobs:
                   {{- toYaml . | nindent 18 }}
                   {{- end }}
+                {{ with $env.cronjobs }}cronjobs:
+                  {{- toYaml . | nindent 18 }}
+                  {{- end }}
                 {{ with $env.secrets }}secrets:
                   {{- toYaml . | nindent 18 }}
                   {{- end }}
--- a/charts/mc-environment/Chart.yaml
+++ b/charts/mc-environment/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: mc-environment
 description: Aggregate multiple mycarrier-helm environments into Argo CD Applications.
 type: application
-version: 0.2.9
+version: 0.2.10
 appVersion: "1.0.0"
--- a/charts/mc-environment/tests/appset_test.yaml
+++ b/charts/mc-environment/tests/appset_test.yaml
@@ -228,3 +228,29 @@ tests:
       - equal:
           path: spec.generators[0].list.elements[0].Values.manualOtelConfig
           value: false
+
+  # mycarrier-helm gained cronjobs; the wrapper builds the subchart values from an explicit
+  # allowlist, so without a passthrough here the key is unreachable for every environment deploy.
+  - it: forwards cronjobs to the subchart values when an environment defines them
+    values:
+      - fixtures/basic-environments.yaml
+    documentSelector:
+      path: metadata.name
+      value: carriers-dev
+    asserts:
+      - equal:
+          path: spec.generators[0].list.elements[0].Values.cronjobs[0].name
+          value: shipments-db-cleanup
+      - equal:
+          path: spec.generators[0].list.elements[0].Values.cronjobs[0].schedule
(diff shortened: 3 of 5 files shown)


charts/mc-environment/templates/appset.yaml
```diff
@@ -36,6 +36,9 @@ spec:
                 {{ with $env.jobs }}jobs:
                   {{- toYaml . | nindent 18 }}
                   {{- end }}
+                {{ with $env.cronjobs }}cronjobs:
+                  {{- toYaml . | nindent 18 }}
+                  {{- end }}
```
hpyvovar-mycarrier, line 41: **🟠 Important — the immediately preceding passthrough shipped with 28 lines of tests; this one ships with none.**

Grounding this on the closest precedent: `03a8dc9` — the last passthrough added to this chart (`manualOtelConfig`) — touched exactly the same three files as this PR (Chart.yaml, appset.yaml, values.schema.json) **plus 28 lines in `charts/mc-environment/tests/appset_test.yaml`**: two tests (opt-out at 206-219, default at 221-230) and a comment at 204-205 whose reasoning applies verbatim here: the wrapper builds subchart values from an explicit allowlist, so without a passthrough the key is unreachable for every environment deploy. This PR is that same change shape with the tests removed.

It matters more than usual because a chart unit test is the **only** pre-merge check that can see this at all: `.github/workflows/main.yml` publishes on merge, `helm-unittest` is the only automated check on the chart, and the render pipeline never validates the result — `mc-environment` renders exactly one file (verified with `--output-dir`: only `appset.yaml`), so kube-linter and kubeconform only ever see ApplicationSets; the CronJob is materialised later by Argo CD at sync time.

Roughly 8 lines: a `cronjobs:` block in `tests/fixtures/basic-environments.yaml` and an `equal` assert on `spec.generators[0].list.elements[0].Values.cronjobs[0].name`, mirroring `03a8dc9`. A `cronjobs: null` case would also cover the new `oneOf [$ref, null]` branch.

hpyvovar-mycarrier (changes requested):

bcarlock-mycarrier (approved):

charts/mc-environment/templates/appset.yaml
pkuzmenko-mycarrier: Fixed in 468b7362 — added helm-unittest coverage mirroring 03a8dc9: a `cronjobs` entry in the test fixture plus two cases — `forwards cronjobs to the subchart values` (asserts the forwarded name+schedule) and `omits cronjobs when absent` (`notExists`, covering the new `oneOf [$ref, null]` branch). `helm unittest charts/mc-environment` → 12 passed.

hpyvovar-mycarrier (approved):

PR #111, opened 2026-08-07
fix(mycarrier-helm): dedupe cronjob annotation keys + cover the securityContext hook

pkuzmenko-mycarrier:
---
Follow-up to #110. That PR merged at `2026-08-07T13:18:03Z`; the CHANGES_REQUESTED review landed at `13:37:37Z`, 19 minutes later, so it could not be edited. Chart **3.5.7 is published on main and currently ships both defects**. This PR fixes them.

## Defect 1 — user annotations could render the same YAML key twice

Review comment: https://github.com/MyCarrier-DevOps/helm-charts/pull/110#discussion_r3736131377

In plain English: `cronjobs[].annotations` were pasted into the rendered manifest as text, *after* the chart's own vault and otel annotations. If your values file set a key the chart also sets, that key appeared **twice** in the same annotations map. kubectl and ArgoCD convert YAML to JSON with last-key-wins, so your value quietly took effect — including for the vault-secrets-webhook directives the chart otherwise pins (`run-as-non-root`, `run-as-user`, `mutate-probes`, `vault-path`). Nothing errored at template, schema, sync, or admission time.

The bigger practical problem is the duplicate key itself. A Trivy 0.61.0 config scan **silently skipped the entire rendered CronJob document** because of it; removing the duplicate made the same document scannable (99 checks). Duplicate keys make downstream scanners blind to everything else in the file.

### What changed

Chart-managed annotations now **always win**. A user key that collides with a chart-managed one is dropped before rendering — it does not override the chart, and it does not render twice. New helper `helm.annotations.userExtra` in `templates/_annotations.tpl` does the filtering; `helm.annotations.cronjobMetadata` exists so `cronjob.yaml` derives its reserved set from the same place it renders from, rather than hardcoding the key in two spots.

Chart-managed keys, scoped per annotation map:
- **Pod template**: all `vault.security.banzaicloud.io/*`, plus `sidecar.opentelemetry.io/inject` and `instrumentation.opentelemetry.io/*` (only when otel autoinstrumentation is on and the language is nodejs/java/python).
- **CronJob metadata**: `argocd.argoproj.io/sync-options`.

`sidecar.istio.io/inject` is not chart-managed and passes through untouched — that is the key the first real consumer actually uses.

### Why a merge and not the suggested `fail`

The review suggested failing the render on collision. I did not adopt that, for two reasons:

1. It is a breaking change for any values file that already carries a colliding key.
2. The reserved set is **context-dependent by design** — otel keys are reserved only when `disableOtelAutoinstrumentation` is false *and* the language matches; `vault-env-from-path` only when `secrets.bulk.path` is set. So a values file renders fine today and starts hard-failing the moment somebody enables an unrelated feature. That break would surface in a future "enable otel" PR, for reasons its author cannot see from their own diff.

A merge emits each key exactly once with an explicit, documented precedence. It fixes the silent override *and* the duplicate-key/scanner problem without planting that landmine. The precedence rule and the full reserved-key list are documented in `values.yaml`.

Blast radius was checked before choosing: across every values file in the org, the only cronjob annotation in use is `sidecar.istio.io/inject: "false"`, which is not chart-managed. **No existing consumer's render changes.**

### The CronJob-metadata surface

`cronjob.yaml` had the identical text-append bug, pre-existing on main. **Fixed here too**, deliberately: it is the same mechanism on the same resource, the Trivy skip is per-document so a metadata-level duplicate blinds the same file, and it reuses the helper already being added. Leaving it would have shipped a precedence rule that holds on only one of the resource's two annotation maps.

### Proof

Same values file, one cronjob whose annotations carry both a benign key and two colliding chart-managed keys:

```
=== BRANCH render, duplicate-key check ===
PASS: no duplicate keys across 1 document(s)          exit=0

=== MAIN render, same values, same check ===
FAIL: Duplicate key found: 'argocd.argoproj.io/sync-options'   exit=1
```

Rendered output, each key exactly once per map:

```yaml
  annotations:                                             # CronJob metadata
    argocd.argoproj.io/sync-options: Prune=true            # reserved here -> chart wins
    sidecar.istio.io/inject: "false"                       # passes through
    vault.security.banzaicloud.io/run-as-non-root: "false"
```
```yaml
          annotations:                                     # pod template
            vault.security.banzaicloud.io/vault-addr: "https://vault.mycarrier.tech"
            vault.security.banzaicloud.io/mutate-probes: "true"
            vault.security.banzaicloud.io/vault-auth-method: azure
            vault.security.banzaicloud.io/vault-path: clusterauth
            vault.security.banzaicloud.io/vault-role: appcluster
            vault.security.banzaicloud.io/run-as-non-root: "true"   # reserved here -> chart wins
            vault.security.banzaicloud.io/run-as-user: "100"
            vault.security.banzaicloud.io/run-as-group: "1000"
            vault.security.banzaicloud.io/vault-agent-env-variables: >-
              [{"Name":"SKIP_CHOWN","Value":"true"},{"Name":"SKIP_SETCAP","Value":"true"}]
            vault.security.banzaicloud.io/vault-env-daemon: "true"
            argocd.argoproj.io/sync-options: Prune=false
            sidecar.istio.io/inject: "false"
```

## Defect 2 — the new securityContext keys shipped untested

Review comment: https://github.com/MyCarrier-DevOps/helm-charts/pull/110#discussion_r3736131389

#110 added exactly one test, covering annotation placement. There was no coverage for `readOnlyRootFilesystem`, `fsGroup`, or `seccompProfile` on cronjobs, and none for the application path those keys had just gone live on.

The sharpest part: the `int64` cast that #110's description headlined as its reason for existing was **protected by nothing**. Confirmed by reproducing it before writing any test — strip `int64` from `_podsecurity.tpl` and `fsGroup: 1000000` renders as `fsGroup: 1e+06`, which the API server rejects, and **all 588 tests still passed**.

### What changed

12 tests added, 588 -> 600, 0 failures:

1. **int64 regression guard** — cronjob `fsGroup: 1000000`. Verified to actually bite by stripping the cast and watching it fail, then restoring:
   ```
   - asserts[1] `equal` fail
       Path:  spec.jobTemplate.spec.template.spec.securityContext.fsGroup
       Expected to equal: 1000000
       Actual: 1e+06
   ```
2. `readOnlyRootFilesystem: true` at container level, plus the negative case.
3. `seccompProfile: {type: RuntimeDefault}` at pod level.
4. Application-level `fsGroup` + `seccompProfile` reaching the Deployment pod spec, plus StatefulSet and Rollout. This pins the contract change — those keys were previously inert and are now honored.
5. Collision tests for Defect 1 on both annotation surfaces.
6. `fsGroup: 0` schema rejection.
7. Snapshot guards for the "default renders byte-identical" claim.

### `fsGroup: 0`

`{{- with $sc.fsGroup }}` treats `0` as falsey, so an explicitly-set `fsGroup: 0` was **silently dropped** — you asked for the root group and got nothing, with no signal. Now `"minimum": 1` in `values.schema.json` (on both the application and cronjob properties) turns it into an explicit rejection. The template guard is unchanged; with the bound in place `0` is unreachable, so the falsiness is no longer a silent-drop path.

The test asserts with `errorPattern` rather than an exact message, because the plugin's validator wording differs from helm's own and differs across plugin versions. The suite was then run under CI's pinned helm-unittest **v1.1.0**, so this is confirmed rather than assumed.

### Snapshot guard — and a `.gitignore` fix it needed

The snapshot tests were initially useless: repo `.gitignore` had a bare `__snapshot__` pattern, so baselines were never tracked. On a fresh CI clone helm-unittest silently regenerates the baseline and passes trivially — a guard that can only agree with itself. Added a narrow negation (`!charts/mycarrier-helm/tests/__snapshot__/`) leaving the broad rule intact, then proved the guard now bites by perturbing `vault-role` and watching both snapshot suites fail.

Worth flagging for anyone verifying locally: **helm-unittest v1.0.0 false-passes `matchSnapshot`** against a deliberately corrupted baseline. CI's pinned **v1.1.0** catches it. Use v1.1.0 for any local snapshot check.

## Defect 3 — job-level `securityContext` was a silent no-op

Found while validating the above, and pulled into this PR because it blocks kyverno policy enforcement.

`jobs.yaml` builds its context as `merge (dict "job" . "ctx" $globalCtx) $`, so per-item values land under the key `job`. But `helm.workloadSecurityContext` only read `.application` and `.cronjob`. Anything you set under `jobs[].securityContext` was therefore **silently discarded** — `fsGroup`, `seccompProfile` and `readOnlyRootFilesystem` never reached the rendered Job, and nothing errored to tell you. On top of that, the `jobs` item schema had no `securityContext` property at all, so there was no schema-level signal either.

That is what blocks kyverno: Jobs rendered by this chart cannot satisfy restricted pod-security policies while the hardening opt-ins do nothing.

Before and after, same job values (`fsGroup: 1000000`, `readOnlyRootFilesystem: true`, `seccompProfile: {type: RuntimeDefault}`):

```yaml
# before — pod-level securityContext on the rendered Job
      securityContext:
        runAsUser: 1000
        runAsGroup: 3000
        runAsNonRoot: true
```
```yaml
# after
      securityContext:
        runAsUser: 1000
        runAsGroup: 3000
        runAsNonRoot: true
        fsGroup: 1000000
        seccompProfile:
          type: RuntimeDefault
```

`_spec_job.tpl` needed no change — it already calls both helpers. The fix is the `.job` branch in the helper plus a `securityContext` block on the jobs item in `values.schema.json`, mirroring cronjobs including `"minimum": 1` on `fsGroup`.

`triggertestengine` was checked and is fine — it passes context key `application`, so it already worked. `job` was the only miswired key.

7 more tests (600 -> 607), including an **opt-in regression guard**: a job with no `securityContext` must render a pod securityContext of exactly `runAsUser`/`runAsGroup`/`runAsNonRoot` and nothing else, so a future change that starts emitting these unconditionally is caught.

Deliberate non-change: `helm.containerSecurityContext` still gates its `addCapabilities` escape hatch on `.application` only. Cronjobs don't get it either, so jobs matching cronjob behaviour is consistent.

## Validation

- `helm lint` — pass on both default and feature values.
- `helm unittest` — 59 suites, **607 tests**, 6 snapshots, **0 failures** (under CI's pinned v1.1.0).
- **Default render byte-identical vs `origin/main` across all six workload templates** that consume the shared helpers: deployment, jobs, rollout, statefulset, triggertestengine, cronjob. 6/6, compared via a clean worktree. Re-verified after the job-context fix, since that change touches the shared helper.

## Notes

- `Chart.yaml` is intentionally left at **3.5.7**. CI bumps it to **3.5.8** on merge.
- Pin sequencing: `workflow-core` currently pins `3.5.6` and `workflow-dev` already pins `3.5.7`. Plan is to move **both to 3.5.8** once this merges, rather than landing anything else on 3.5.7.

## Review round two

All 15 review threads addressed. Summary of what changed beyond the original scope:

- **Annotation filtering now covers every surface**, not just CronJob: the three application pod templates, Job resource metadata, and the Job pod template. `istio`/`gateway` keys are reserved on application surfaces too — `applications[].istioDisabled` and `networking.istio.enabled` are the supported knobs, and leaving the annotation open kept a live duplicate key.
- **`jobs[].annotations` now reach the Job pod template.** Previously impossible. Chart Jobs are ArgoCD PreSync/PostSync hooks, so a never-exiting Istio sidecar blocks the whole application sync.
- **Schema deduplicated and tightened**: one `#/definitions/workloadSecurityContext`, `additionalProperties: false`, `seccompProfile` requiring `type` against the three legal Kubernetes values, `addCapabilities` declared.
- **`_podsecurity.tpl`**: exclusive `else if` chain; presence-based `hasKey` guards.
- **Dead `| default dict` removed** from all seven reserved-set sites, replaced by one fail-fast guard that fires only on chart-authored parse failures.
- **Docs corrected**: reserved set is exact keys not a `vault.*` prefix; the kyverno `restrict-seccomp` note is now per-cluster accurate (Enforce on the development cluster accepting `RuntimeDefault`; Audit elsewhere accepting only `runtime/default`); `fsGroup >= 1` attributed to org policy, since Pod Security Standards does not constrain `fsGroup` at all.

Two suggestions were declined with evidence (the dict-refactor remedy, which breaks byte-identity; and `@param` entries, which the README generator rejects for array-item paths), and the kyverno finding was corrected against live cluster state. Details in the individual threads.

**Validation**: 59 suites, **623 tests** (+16), 6 snapshots, 0 failures. Default render byte-identical to main across all six workload templates. A values file engineered to collide on every annotation surface renders **zero duplicate YAML keys** on this branch; the same input on `main` produces three.

### DIFF
--- a/charts/mycarrier-helm/templates/_annotations.tpl
+++ b/charts/mycarrier-helm/templates/_annotations.tpl
@@ -5,6 +5,18 @@ as a duplicate YAML key. Returns "" when nothing survives, so callers can guard 
 */}}
 {{- define "helm.annotations.userExtra" -}}
 {{- $reserved := .reserved | default dict -}}
+{{- /* Fail fast on a CHART-AUTHORED parse failure only. `fromYaml` returns a non-empty
+     {"Error": "..."} map (not nil) when one of the chart's own annotation snippets fails
+     to parse as YAML, so a plain `hasKey $reserved "Error"` here can only be tripped by
+     chart-managed content: $reserved is built exclusively from `include ... | fromYaml`
+     calls over chart templates (helm.annotations.vault/istio/gateway, helm.otel.annotations,
+     helm.annotations.cronjobMetadata/jobMetadata, the argocd reserved-key dict) — user
+     annotations are never merged into $reserved, only compared against it below — so a
+     user annotation literally keyed "Error" cannot land in $reserved and cannot trigger
+     this guard. Do NOT generalise this into failing on user input: the chart's policy is
+     that a colliding user key is silently dropped (see the file header comment), never
+     fatal. */ -}}
+{{- if hasKey $reserved "Error" }}{{- fail (printf "helm.annotations.userExtra: chart-managed annotations failed to parse: %v" (get $reserved "Error")) }}{{- end }}
 {{- $out := dict -}}
 {{- range $k, $v := (.user | default dict) -}}
 {{- if not (hasKey $reserved $k) -}}
@@ -24,4 +36,14 @@ rather than hardcoding the key twice and risking drift.
 */}}
 {{- define "helm.annotations.cronjobMetadata" -}}
 argocd.argoproj.io/sync-options: Prune=true
+{{- end -}}
+
+{{/*
+The chart-managed Job *metadata* annotations (as opposed to the pod-template annotations
+covered by helm.annotations.vault / helm.otel.annotations). Kept as its own define so
+jobs.yaml can derive its reserved-key set from the exact same source it renders from,
+rather than hardcoding the key twice and risking drift.
+*/}}
+{{- define "helm.annotations.jobMetadata" -}}
+argocd.argoproj.io/sync-options: Force=true,Replace=true
 {{- end -}}
--- a/charts/mycarrier-helm/templates/_annotations.tpl
+++ b/charts/mycarrier-helm/templates/_annotations.tpl
@@ -1,0 +1,27 @@
+{{/*
+Emit the user-supplied annotations that do NOT collide with chart-managed keys.
+Chart-managed keys always win: a colliding user key is dropped rather than emitted
+as a duplicate YAML key. Returns "" when nothing survives, so callers can guard with `if`.
+*/}}
+{{- define "helm.annotations.userExtra" -}}
+{{- $reserved := .reserved | default dict -}}
+{{- $out := dict -}}
+{{- range $k, $v := (.user | default dict) -}}
+{{- if not (hasKey $reserved $k) -}}
+{{- $_ := set $out $k $v -}}
+{{- end -}}
+{{- end -}}
+{{- if $out -}}
+{{- toYaml $out -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+The chart-managed CronJob *metadata* annotations (as opposed to the pod-template annotations
+covered by helm.annotations.vault / helm.otel.annotations). Kept as its own define so
+cronjob.yaml can derive its reserved-key set from the exact same source it renders from,
+rather than hardcoding the key twice and risking drift.
+*/}}
+{{- define "helm.annotations.cronjobMetadata" -}}
+argocd.argoproj.io/sync-options: Prune=true
+{{- end -}}
--- a/charts/mycarrier-helm/templates/_podsecurity.tpl
+++ b/charts/mycarrier-helm/templates/_podsecurity.tpl
@@ -3,16 +3,17 @@ Resolve the workload-item securityContext dict (opt-in hardening block) regardle
 whether the caller context is an application (.application.securityContext), a cronjob
 item (.cronjob.securityContext), a job item (.job.securityContext), or neither (defaults
 to an empty dict so all lookups below are no-ops and rendered output is unchanged).
+The three sources are mutually exclusive (application -> cronjob -> job precedence,
+matching the exclusive if/else-if chain used in _labels.tpl) since a single caller
+context only ever carries one of .application/.cronjob/.job.
 */}}
 {{- define "helm.workloadSecurityContext" -}}
 {{- $sc := dict -}}
 {{- if and .application .application.securityContext -}}
 {{- $sc = .application.securityContext -}}
-{{- end -}}
-{{- if and .cronjob .cronjob.securityContext -}}
+{{- else if and .cronjob .cronjob.securityContext -}}
 {{- $sc = .cronjob.securityContext -}}
-{{- end -}}
-{{- if and .job .job.securityContext -}}
+{{- else if and .job .job.securityContext -}}
 {{- $sc = .job.securityContext -}}
 {{- end -}}
 {{- $sc | toJson -}}
@@ -25,12 +26,12 @@ securityContext:
   runAsUser: 1000
   runAsGroup: 3000
   runAsNonRoot: true
-{{- with $sc.fsGroup }}
-  fsGroup: {{ int64 . }}
+{{- if hasKey $sc "fsGroup" }}
+  fsGroup: {{ int64 $sc.fsGroup }}
 {{- end }}
-{{- with $sc.seccompProfile }}
+{{- if hasKey $sc "seccompProfile" }}
   seccompProfile:
-{{ toYaml . | indent 4 }}
+{{ toYaml $sc.seccompProfile | indent 4 }}
 {{- end }}
 {{- end }}
 {{- end -}}
--- a/charts/mycarrier-helm/templates/_podsecurity.tpl
+++ b/charts/mycarrier-helm/templates/_podsecurity.tpl
@@ -1,8 +1,8 @@
 {{/*
 Resolve the workload-item securityContext dict (opt-in hardening block) regardless of
 whether the caller context is an application (.application.securityContext), a cronjob
-item (.cronjob.securityContext), or neither (defaults to an empty dict so all lookups
-below are no-ops and rendered output is unchanged).
+item (.cronjob.securityContext), a job item (.job.securityContext), or neither (defaults
+to an empty dict so all lookups below are no-ops and rendered output is unchanged).
 */}}
 {{- define "helm.workloadSecurityContext" -}}
 {{- $sc := dict -}}
@@ -11,6 +11,9 @@ below are no-ops and rendered output is unchanged).
 {{- end -}}
 {{- if and .cronjob .cronjob.securityContext -}}
 {{- $sc = .cronjob.securityContext -}}
+{{- end -}}
+{{- if and .job .job.securityContext -}}
+{{- $sc = .job.securityContext -}}
 {{- end -}}
 {{- $sc | toJson -}}
 {{- end -}}
--- a/charts/mycarrier-helm/templates/_spec_cronjob.tpl
+++ b/charts/mycarrier-helm/templates/_spec_cronjob.tpl
@@ -48,8 +48,10 @@ jobTemplate:
         annotations:
           {{ include "helm.annotations.vault" . | indent 10 | trim }}
           {{ include "helm.otel.annotations" . | indent 10 | trim }}
-          {{- with .cronjob.annotations }}
-          {{ toYaml . | indent 10 | trim }}
+          {{- $reserved := merge (dict) (include "helm.annotations.vault" $ | fromYaml | default dict) (include "helm.otel.annotations" $ | fromYaml | default dict) }}
+          {{- $extra := include "helm.annotations.userExtra" (dict "reserved" $reserved "user" $.cronjob.annotations) }}
+          {{- if $extra }}
+          {{ $extra | indent 10 | trim }}
           {{- end }}
       spec:
         {{ include "helm.podSecurityContext" . | indent 8 | trim }}
--- a/charts/mycarrier-helm/templates/_spec_cronjob.tpl
+++ b/charts/mycarrier-helm/templates/_spec_cronjob.tpl
@@ -48,7 +48,7 @@ jobTemplate:
         annotations:
           {{ include "helm.annotations.vault" . | indent 10 | trim }}
           {{ include "helm.otel.annotations" . | indent 10 | trim }}
-          {{- $reserved := merge (dict) (include "helm.annotations.vault" $ | fromYaml | default dict) (include "helm.otel.annotations" $ | fromYaml | default dict) }}
+          {{- $reserved := merge (dict) (include "helm.annotations.vault" $ | fromYaml) (include "helm.otel.annotations" $ | fromYaml) }}
           {{- $extra := include "helm.annotations.userExtra" (dict "reserved" $reserved "user" $.cronjob.annotations) }}
           {{- if $extra }}
           {{ $extra | indent 10 | trim }}
--- a/charts/mycarrier-helm/templates/cronjob.yaml
+++ b/charts/mycarrier-helm/templates/cronjob.yaml
@@ -20,9 +20,11 @@ metadata:
     {{ toYaml .labels | indent 4 | trim }}
     {{- end }}
   annotations:
-    argocd.argoproj.io/sync-options: Prune=true
-  {{- with .annotations }}
-    {{ toYaml . | indent 4 | trim }}
+    {{ include "helm.annotations.cronjobMetadata" . | indent 4 | trim }}
+  {{- $reserved := include "helm.annotations.cronjobMetadata" . | fromYaml | default dict }}
+  {{- $extra := include "helm.annotations.userExtra" (dict "reserved" $reserved "user" .annotations) }}
+  {{- if $extra }}
+    {{ $extra | indent 4 | trim }}
   {{- end }}
 spec:
   {{ include "helm.specs.cronjob" $rootContext | indent 2 | trim }}
--- a/charts/mycarrier-helm/templates/cronjob.yaml
+++ b/charts/mycarrier-helm/templates/cronjob.yaml
@@ -21,7 +21,7 @@ metadata:
     {{- end }}
   annotations:
     {{ include "helm.annotations.cronjobMetadata" . | indent 4 | trim }}
-  {{- $reserved := include "helm.annotations.cronjobMetadata" . | fromYaml | default dict }}
+  {{- $reserved := include "helm.annotations.cronjobMetadata" . | fromYaml }}
   {{- $extra := include "helm.annotations.userExtra" (dict "reserved" $reserved "user" .annotations) }}
   {{- if $extra }}
     {{ $extra | indent 4 | trim }}
--- a/charts/mycarrier-helm/tests/security_context_test.yaml
+++ b/charts/mycarrier-helm/tests/security_context_test.yaml
@@ -151,3 +151,45 @@ tests:
           value:
             - NET_BIND_SERVICE
             - SYS_TIME
+
+  - it: should honor application-level securityContext fsGroup and seccompProfile on the pod spec
+    set:
+      environment.name: "dev"
+      applications.app-level-sc-app:
+        enabled: true
+        deploymentType: deployment
+        image:
+          registry: "test-registry"
+          repository: "app-level-sc-app"
+          tag: "1.0.0"
+        securityContext:
+          fsGroup: 1000000
+          seccompProfile:
+            type: RuntimeDefault
+    asserts:
+      - isKind:
+          of: Deployment
+      - equal:
+          path: spec.template.spec.securityContext.fsGroup
+          value: 1000000
+      - equal:
+          path: spec.template.spec.securityContext.seccompProfile.type
+          value: RuntimeDefault
+
+  - it: should render pod-template annotations and pod securityContext byte-identical to snapshot (regression guard)
+    set:
+      environment.name: "dev"
+      applications.snapshot-guard-app:
+        enabled: true
+        deploymentType: deployment
+        image:
+          registry: "test-registry"
+          repository: "snapshot-guard-app"
+          tag: "1.0.0"
+    asserts:
+      - isKind:
+          of: Deployment
+      - matchSnapshot:
+          path: spec.template.metadata.annotations
+      - matchSnapshot:
+          path: spec.template.spec.securityContext
--- a/charts/mycarrier-helm/tests/security_context_test.yaml
+++ b/charts/mycarrier-helm/tests/security_context_test.yaml
@@ -193,3 +193,63 @@ tests:
           path: spec.template.metadata.annotations
       - matchSnapshot:
           path: spec.template.spec.securityContext
+
+  - it: should reject fsGroup 0 via the values schema on the application securityContext (fsGroup 0 is the root group and is disallowed)
+    set:
+      environment.name: "dev"
+      applications.fsgroup-zero-app:
+        enabled: true
+        deploymentType: deployment
+        image:
+          registry: "test-registry"
+          repository: "fsgroup-zero-app"
+          tag: "1.0.0"
+        securityContext:
+          fsGroup: 0
+    asserts:
+      # Loosely asserted for CI-version robustness (see the equivalent job/cronjob tests):
+      # local unittest v1.0.0 emits "Must be greater than or equal to 1"; CI pins v1.1.0 whose
+      # exact wording is unconfirmed - jsonschema's "minimum: got 0, want 1" is one known form.
+      - failedTemplate:
+          errorPattern: "fsGroup"
+      - failedTemplate:
+          errorPattern: "(?i)(greater than or equal to 1|minimum|want 1|>= *1)"
+
+  - it: should render the smallest legal fsGroup (1) on an application as a presence guard boundary, not a falsey-guard boundary
+    set:
+      environment.name: "dev"
+      applications.fsgroup-one-app:
+        enabled: true
+        deploymentType: deployment
+        image:
+          registry: "test-registry"
+          repository: "fsgroup-one-app"
+          tag: "1.0.0"
+        securityContext:
+          fsGroup: 1
+    asserts:
+      - isKind:
+          of: Deployment
+      - equal:
+          path: spec.template.spec.securityContext.fsGroup
+          value: 1
+
+  - it: should render pod securityContext with only the default fields when no application securityContext is specified (opt-in regression guard)
+    set:
+      environment.name: "dev"
+      applications.no-sc-app:
+        enabled: true
+        deploymentType: deployment
+        image:
+          registry: "test-registry"
+          repository: "no-sc-app"
+          tag: "1.0.0"
+    asserts:
+      - isKind:
+          of: Deployment
+      - equal:
+          path: spec.template.spec.securityContext
+          value:
+            runAsUser: 1000
+            runAsGroup: 3000
+            runAsNonRoot: true
--- a/charts/mycarrier-helm/values.schema.json
+++ b/charts/mycarrier-helm/values.schema.json
@@ -606,7 +606,8 @@
               },
               "fsGroup": {
                 "type": "integer",
-                "description": "Pod-level fsGroup (opt-in; unset by default)"
+                "description": "Pod-level fsGroup (opt-in; unset by default; must be >= 1 — fsGroup 0 is the root group and is rejected)",
+                "minimum": 1
               },
               "seccompProfile": {
                 "type": "object",
@@ -1988,7 +1989,8 @@
               },
               "fsGroup": {
                 "type": "integer",
-                "description": "Pod-level fsGroup (opt-in; unset by default)"
+                "description": "Pod-level fsGroup (opt-in; unset by default; must be >= 1 — fsGroup 0 is the root group and is rejected)",
+                "minimum": 1
               },
               "seccompProfile": {
                 "type": "object",
--- a/charts/mycarrier-helm/values.schema.json
+++ b/charts/mycarrier-helm/values.schema.json
@@ -1726,6 +1726,26 @@
                 }
               }
             }
+          },
+          "securityContext": {
+            "type": "object",
+            "description": "Opt-in hardened security context for this job",
+            "properties": {
+              "readOnlyRootFilesystem": {
+                "type": "boolean",
+                "description": "Container-level readOnlyRootFilesystem (opt-in; unset defaults to false)",
+                "default": false
+              },
+              "fsGroup": {
+                "type": "integer",
+                "description": "Pod-level fsGroup (opt-in; unset by default; must be >= 1 — fsGroup 0 is the root group and is rejected)",
+                "minimum": 1
+              },
+              "seccompProfile": {
+                "type": "object",
+                "description": "Pod-level seccompProfile (opt-in; unset by default)"
+              }
+            }
           }
         }
       }
--- a/charts/mycarrier-helm/values.schema.json
+++ b/charts/mycarrier-helm/values.schema.json
@@ -169,6 +169,18 @@
               }
             ]
           },
+          "istioDisabled": {
+            "oneOf": [
+              {
+                "type": "boolean",
+                "description": "Opt this application's pod template out of Istio sidecar injection and drop the mycarrier.io/gateway annotation (read by _helpers.tpl helm.annotations.istio / helm.annotations.gateway). Not to be confused with service.istioDisabled, which is a separate per-service knob.",
+                "default": false
+              },
+              {
+                "type": "null"
+              }
+            ]
+          },
           "staticHostname": {
             "oneOf": [
               {
@@ -598,22 +610,7 @@
             }
           },
           "securityContext": {
-            "type": "object",
-            "properties": {
-              "readOnlyRootFilesystem": {
-                "type": "boolean",
-                "default": false
-              },
-              "fsGroup": {
-                "type": "integer",
-                "description": "Pod-level fsGroup (opt-in; unset by default; must be >= 1 — fsGroup 0 is the root group and is rejected)",
-                "minimum": 1
-              },
-              "seccompProfile": {
-                "type": "object",
-                "description": "Pod-level seccompProfile (opt-in; unset by default)"
-              }
-            }
+            "$ref": "#/definitions/workloadSecurityContext"
           },
           "replicas": {
             "type": "integer",
@@ -1728,24 +1725,7 @@
             }
           },
           "securityContext": {
-            "type": "object",
-            "description": "Opt-in hardened security context for this job",
-            "properties": {
-              "readOnlyRootFilesystem": {
-                "type": "boolean",
-                "description": "Container-level readOnlyRootFilesystem (opt-in; unset defaults to false)",
-                "default": false
-              },
-              "fsGroup": {
-                "type": "integer",
-                "description": "Pod-level fsGroup (opt-in; unset by default; must be >= 1 — fsGroup 0 is the root group and is rejected)",
-                "minimum": 1
-              },
-              "seccompProfile": {
-                "type": "object",
-                "description": "Pod-level seccompProfile (opt-in; unset by default)"
-              }
-            }
+            "$ref": "#/definitions/workloadSecurityContext"
           }
         }
       }
@@ -1999,24 +1979,7 @@
             }
           },
           "securityContext": {
-            "type": "object",
-            "description": "Opt-in hardened security context for this cronjob",
-            "properties": {
-              "readOnlyRootFilesystem": {
-                "type": "boolean",
-                "description": "Container-level readOnlyRootFilesystem (opt-in; unset defaults to false)",
-                "default": false
-              },
-              "fsGroup": {
-                "type": "integer",
-                "description": "Pod-level fsGroup (opt-in; unset by default; must be >= 1 — fsGroup 0 is the root group and is rejected)",
-                "minimum": 1
-              },
-              "seccompProfile": {
-                "type": "object",
-                "description": "Pod-level seccompProfile (opt-in; unset by default)"
-              }
-            }
+            "$ref": "#/definitions/workloadSecurityContext"
           }
         }
       }
@@ -2428,6 +2391,48 @@
                 }
               }
             }
+          }
+        }
+      }
+    }
+  },
+  "definitions": {
+    "workloadSecurityContext": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "readOnlyRootFilesystem": {
+          "type": "boolean",
+          "description": "Container-level readOnlyRootFilesystem (opt-in; unset defaults to false)",
+          "default": false
+        },
+        "fsGroup": {
+          "type": "integer",
+          "description": "Pod-level fsGroup (opt-in; unset by default). Must be empty or greater than zero per MyCarrier-DevOps/kyverno-policies `require-non-root-groups` rule `check-fsGroup` — fsGroup 0 is the root group.",
+          "minimum": 1
+        },
+        "seccompProfile": {
+          "type": "object",
+          "description": "Pod-level seccompProfile (opt-in; unset by default)",
+          "properties": {
+            "type": {
+              "type": "string",
+              "description": "SeccompProfile type, per the Kubernetes SeccompProfile.type API field",
+              "enum": ["RuntimeDefault", "Localhost", "Unconfined"]
+            },
+            "localhostProfile": {
+              "type": "string",
+              "description": "Path to a pre-configured profile on the node, relative to the kubelet's configured seccomp profile location. Only used when type is Localhost."
+            }
+          },
+          "required": ["type"],
+          "additionalProperties": false
+        },
+        "addCapabilities": {
+          "type": "array",
+          "description": "Linux capabilities to add back on top of the drop-all default. All capabilities are dropped by default; list ones here to add them back. Only honoured on the application (Deployment) path by _podsecurity.tpl — jobs and cronjobs do not read this field.",
+          "items": {
+            "type": "string"
           }
         }
       }
--- a/charts/mycarrier-helm/values.yaml
+++ b/charts/mycarrier-helm/values.yaml
@@ -109,12 +109,44 @@ deployment: deployment
 ## **Container**: image.registry, image.repository, image.tag, pullPolicy, pullSecret
 ## **Resources**: resources.requests.cpu/memory, resources.limits.cpu/memory
 ## **Health Probes**: probes.enableLiveness, probes.enableReadiness, probes.enableStartup (all default true; set to false to disable), custom probe configurations
-## **Security Context**: securityContext.readOnlyRootFilesystem (default false), securityContext.addCapabilities (default []; all capabilities are dropped by default, list specific capabilities here to add them back), securityContext.fsGroup (opt-in pod-level fsGroup; previously ignored/silently dropped, now honored; must be >= 1 — fsGroup 0 is the root group and is rejected by the values schema), securityContext.seccompProfile (opt-in pod-level seccompProfile; previously ignored/silently dropped, now honored)
+## **Security Context**: securityContext.readOnlyRootFilesystem (default false), securityContext.addCapabilities (default []; all capabilities are dropped by default, list specific capabilities here to add them back), securityContext.fsGroup (opt-in pod-level fsGroup; previously ignored/silently dropped, now honored; must be >= 1 — fsGroup 0 is the root group and is rejected by the values schema; the >= 1 bound is ORG POLICY enforced by MyCarrier-DevOps/kyverno-policies `require-non-root-groups` rule `check-fsGroup` ("must be empty or greater than zero"), not an upstream Pod Security Standards control — the PSS spec contains zero occurrences of fsGroup/runAsGroup/supplementalGroups; those were PodSecurityPolicy fields, and PSP is retired), securityContext.seccompProfile (opt-in pod-level seccompProfile; previously ignored/silently dropped, now honored)
 ## **Service**: service.type, service.ports, service.annotations, service.timeout, service.retryOn
 ## **Networking**: networking.ingress.type, networking.istio.enabled/hosts/routes/corsPolicy/allowedEndpoints
 ## **Autoscaling**: HPA (autoscaling.minReplicas/maxReplicas/targetCPU/targetMemory), VPA (vpa.enabled/updateMode/resources)
 ## **Monitoring**: serviceMonitor.enabled/path/interval/scheme
 ## **Testing**: testtrigger configuration for automated test jobs
+##
+## **Annotations — reserved keys (application pod template)**: `applications[].annotations` render
+## onto the pod template for Deployment / StatefulSet / Rollout applications. Chart-managed
+## annotations always win: a user-supplied key that collides with a chart-managed key is DROPPED
+## from `annotations` before rendering — it does not override the chart's value and does not render
+## twice. The reserved set is an EXACT-KEY match (`hasKey`), not a `vault.security.banzaicloud.io/*`
+## prefix match:
+##   - vault keys: `vault-addr`, `vault-auth-method`, `vault-path`, `vault-role`,
+##     `vault-agent-env-variables`, `mutate-probes`, `run-as-non-root`, `run-as-user`,
+##     `run-as-group`, plus `vault-env-daemon` (only when `secrets` is set) and
+##     `vault-env-from-path` (only when `secrets.bulk.path` is set)
+##   - otel keys: `sidecar.opentelemetry.io/inject`, `instrumentation.opentelemetry.io/container-names`,
+##     and `instrumentation.opentelemetry.io/inject-<language>` for the ONE resolved language only.
+##     Rendered only when `manualOtelConfig` is not true AND `disableOtelAutoinstrumentation: false`
+##     AND the language is nodejs/java/python (verified call chain: `helm.otel.annotations` ->
+##     `helm.otel.autoinstrumentation.enabled` -> `helm.otel.chartManaged` -> `manualOtelConfig`)
+##   - istio keys: `sidecar.istio.io/inject`, `proxy.istio.io/config`
+##   - gateway key: `mycarrier.io/gateway`
+##
+## Any key NOT in this reserved set is applied to the pod as-is — including other
+## `vault.security.banzaicloud.io/*` keys the chart does not set. Security-relevant keys verified to
+## pass through unmanaged: `vault-skip-verify` (disables TLS verification to Vault), `mutate: skip`
+## (disables secret injection entirely), `vault-namespace`, `vault-serviceaccount`.
+## `psp-allow-privilege-escalation` also renders through but is not documented in the current
+## bank-vaults reference.
+##
+## **Istio override is CLOSED on application surfaces**: `sidecar.istio.io/inject`,
+## `proxy.istio.io/config` and `mycarrier.io/gateway` set in `applications[].annotations` are DROPPED
+## — the chart's own computed values always win. The supported per-workload knobs are
+## `applications[].istioDisabled`, `applications[].networking.istio.enabled`, and
+## `applications[].service.istioDisabled`. This closure does NOT apply to `cronjobs[]`/`jobs[]` — see
+## the Kubernetes Jobs / CronJobs sections below.
 
 ## @param applications [object] Map of application definitions where the key is the application name
 
@@ -593,14 +625,49 @@ applications: {}
 ##
 ## **Security Context (opt-in)**: securityContext.readOnlyRootFilesystem (default false; sets the
 ## container's readOnlyRootFilesystem), securityContext.fsGroup (unset by default; sets the pod-level
-## fsGroup; must be >= 1 — fsGroup 0 is the root group and is rejected by the values schema),
-## securityContext.seccompProfile (unset by default; sets the pod-level seccompProfile, e.g.
-## {type: RuntimeDefault}). All three are opt-in per job item — omitting securityContext entirely
-## preserves the chart's existing default rendering.
+## fsGroup; must be >= 1 — fsGroup 0 is the root group and is rejected by the values schema; the >= 1
+## bound is ORG POLICY enforced by MyCarrier-DevOps/kyverno-policies `require-non-root-groups` rule
+## `check-fsGroup` ("must be empty or greater than zero"), not an upstream Pod Security Standards
+## control — the PSS spec contains zero occurrences of fsGroup/runAsGroup/supplementalGroups; those
+## were PodSecurityPolicy fields, and PSP is retired), securityContext.seccompProfile (unset by
+## default; sets the pod-level seccompProfile, e.g. {type: RuntimeDefault}). All three are opt-in per
+## job item — omitting securityContext entirely preserves the chart's existing default rendering.
 ##
-## IMPORTANT: the cluster's kyverno `restrict-seccomp` policy is Enforce and only allows
-## seccompProfile.type `RuntimeDefault` (or `runtime/default`) — `Localhost` renders fine from this
-## chart but is hard-rejected at admission. Stick to `RuntimeDefault`.
+## the `restrict-seccomp` ClusterPolicy differs per cluster, because each app cluster syncs its own
+## branch of MyCarrier-DevOps/kyverno-policies:
+##   - development cluster: validationFailureAction Enforce; pattern accepts
+##     `runtime/default | RuntimeDefault`, so `RuntimeDefault` passes and IS enforced at admission.
+##   - data and production-csp clusters: validationFailureAction Audit (report-only — violations
+##     appear in PolicyReport objects, nothing is blocked); pattern accepts only the literal
+##     `runtime/default`, so `RuntimeDefault` records a PolicyReport entry but is not rejected.
+## Use `RuntimeDefault`: it is the correct value for the Kubernetes SeccompProfile.type field
+## (`runtime/default` is the pre-1.25 annotation form and is not a legal field value), and Pod
+## Security Standards `restricted` permits `RuntimeDefault` and `Localhost`, forbidding only
+## `Unconfined`.
+##
+## **Annotations**: `jobs[].annotations` render onto BOTH the Job resource metadata and the Job pod
+## template (filtered), mirroring `cronjobs[].annotations`. This matters because chart Jobs are
+## ArgoCD PreSync/PostSync hooks — a never-exiting Istio sidecar on a Job pod blocks the entire
+## application sync, not just one Job run — so `sidecar.istio.io/inject: "false"` set via
+## `jobs[].annotations` is the documented way to stop that. The Job pod template emits no istio
+## annotations of its own, so this key is not chart-managed there and passes through unaffected
+## (unlike the application pod template, where it is chart-managed and closed to overrides).
+##
+## Chart-managed annotations always win; a colliding user-supplied key is DROPPED from `annotations`
+## before rendering. The reserved set is an EXACT-KEY match (`hasKey`), not a
+## `vault.security.banzaicloud.io/*` prefix match:
+##   - Job resource metadata: `argocd.argoproj.io/sync-options`
+##   - Job pod template: vault keys (`vault-addr`, `vault-auth-method`, `vault-path`, `vault-role`,
+##     `vault-agent-env-variables`, `mutate-probes`, `run-as-non-root`, `run-as-user`, `run-as-group`,
+##     plus `vault-env-daemon` when `secrets` is set and `vault-env-from-path` when
+##     `secrets.bulk.path` is set) and otel keys (`sidecar.opentelemetry.io/inject`,
+##     `instrumentation.opentelemetry.io/container-names`,
+##     `instrumentation.opentelemetry.io/inject-<language>` — see the Applications section above for
+##     the full otel gating conditions), plus `argocd.argoproj.io/sync-options`,
+##     `argocd.argoproj.io/hook-delete-policy`, `argocd.argoproj.io/sync-wave`, and
+##     `argocd.argoproj.io/hook`. The four argocd keys are reserved UNCONDITIONALLY, even though
+##     sync-wave/hook only render when `.timing` is set — this keeps the reserved set independent of
+##     `.timing`, so enabling a hook later cannot silently change which user keys survive.
 ##
 ## See the example configuration below for complete job specification details.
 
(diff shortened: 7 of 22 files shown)


charts/mycarrier-helm/templates/_annotations.tpl
```diff
@@ -0,0 +1,27 @@
...
+{{- $out := dict -}}
+{{- range $k, $v := (.user | default dict) -}}
+{{- if not (hasKey $reserved $k) -}}
+{{- $_ := set $out $k $v -}}
+{{- end -}}
+{{- end -}}
+{{- if $out -}}
+{{- toYaml $out -}}
+{{- end -}}
+{{- end -}}
```
smunukutla-mycarrier, line 29: **🟠 important · security · Security-IaC**

### `userExtra` is wired into CronJob only — the "chart-managed annotations always win" guarantee is false on 4 of 5 surfaces

The helper is correct and genuinely fixes the CronJob surface (verified base `117ed8ac` → HEAD). It is wired into exactly two call sites: `cronjob.yaml:25` and `_spec_cronjob.tpl:52`. Four sibling surfaces still text-append user annotations after the chart-managed block with no filtering:

```
_spec_deployment.tpl:72-74   {{ with .application.annotations }}{{ toYaml . | indent 6 | trim }}
_spec_statefulset.tpl:34-36  same pattern
_spec_rollout.tpl:82-84      same pattern
jobs.yaml:24-26              {{ with .annotations }} after argocd.argoproj.io/sync-options
```

**Scope correction, stated up front:** this is *not* a privilege-escalation finding. There is no boundary between "user annotations" and "chart values" — the same principal writes both, and `disableSecurity: true` validates against the root schema and strips every securityContext block via `_podsecurity.tpl:22`/`:39`. The finding is about the integrity of a chart-declared invariant, and about documentation this PR adds asserting a guarantee the code does not provide.

**Reproduction** (exact invocation, so line numbers are stable):

```
helm template t charts/mycarrier-helm -f collide-app.yaml \
  --show-only templates/deployment.yaml
```

with `applications.api.annotations` setting `run-as-non-root: "false"`, `run-as-user: "0"`, `vault-addr: "https://evil.example.com"`. Chart-managed keys land at lines 62/67/68; the user's duplicates at 78/79/80. StatefulSet and Rollout reproduce identically (3 occurrences of `run-as-non-root` each); `jobs.yaml` yields a duplicate `argocd.argoproj.io/sync-options`.

**The consequence is a disjunction, and both branches are defects.** Tested directly against `sigs.k8s.io/yaml` v1.4.0 — the decoder kubectl, client-go and controller-runtime use:

```
Unmarshal (non-strict)  err = <nil>
  EFFECTIVE vault.security.banzaicloud.io/run-as-non-root = "false"
  EFFECTIVE vault.security.banzaicloud.io/run-as-user     = "0"
UnmarshalStrict         err = error converting YAML to JSON: yaml: unmarshal errors:
  line 8: key "vault.security.banzaicloud.io/run-as-non-root" already set in map
  line 9: key "vault.security.banzaicloud.io/run-as-user" already set in map
```

On a non-strict path the chart's pinned value is silently replaced; on a strict path the manifest is rejected. Per the [bank-vaults reference](https://bank-vaults.dev/docs/mutating-webhook/annotations/), `run-as-non-root` defaults to `"false"` and `run-as-user` to `"0"`, so the non-strict outcome returns injected vault-env containers to the webhook's root defaults.

**Scanner-blinding is real but latent for this org**, and severity does not rest on it: Trivy 0.72.0 on an isolated single-document Deployment render with the duplicate reports `Detected config files num=0` and zero results (exit 0); the same render without the collision reports 6 misconfigurations including `KSV-0014 HIGH`. However, no trivy/checkov/kubesec/polaris step exists in this repo's CI, so nothing consumes that signal today. Forward-looking only.

**Which keys stay overridable — this is why the fix is not a copy-paste.** The three pod templates render **four** chart-managed helpers, not two. Reserve only `helm.annotations.vault` and `helm.otel.annotations`. Do **not** reserve `helm.annotations.istio` (`_helpers.tpl:202-209`, always emits `sidecar.istio.io/inject` and `proxy.istio.io/config`) or `helm.annotations.gateway` (`_helpers.tpl:218`, emits `mycarrier.io/gateway`). Those are routing/observability toggles, not security controls; the annotation override is today the only per-workload Istio escape hatch, and the failure mode of a user override there is an un-meshed pod, not a weakened secret-injection posture. Reserving them would remove capability with no replacement.

For `jobs.yaml` the minimum correct fix is simply to stop emitting a duplicate key; the precedence question for `argocd.argoproj.io/sync-options` is raised separately on `_annotations.tpl:1-5`.

Pre-existing rather than introduced here, so not blocking. But this PR ships documentation asserting "Chart-managed annotations always win" while that holds only for CronJob. Either extend the helper to the four surfaces with the scoping above, or scope the documentation claim to CronJob. Shipping the general claim with the narrow implementation is the worst of the three.

charts/mycarrier-helm/values.yaml
```diff
@@ -643,18 +662,32 @@ jobs: []
...
+## CronJob's pods out of Istio sidecar injection in istio-injection=enabled namespaces (a never-exiting
+## sidecar container otherwise hangs the Job). `sidecar.istio.io/inject` is not chart-managed and
+## passes through unaffected.
+##
+## Chart-managed annotations always win. A user-supplied key that collides with a chart-managed
+## annotation is DROPPED from `annotations` before rendering — it will not override the chart's value,
+## and it will not render twice. Chart-managed keys are: all `vault.security.banzaicloud.io/*` (pod
+## template), the otel keys `sidecar.opentelemetry.io/inject` and `instrumentation.opentelemetry.io/*`
+## (pod template, only when otel autoinstrumentation is enabled AND the app language is
+## nodejs/java/python), and `argocd.argoproj.io/sync-options` (CronJob metadata).
```
smunukutla-mycarrier: **🟠 important · security · Security-IaC**

### Reserved set is ~10 exact keys, not the documented "all `vault.security.banzaicloud.io/*`" prefix

`helm.annotations.userExtra` filters with `hasKey $reserved $k` (`_annotations.tpl:10`) — an exact-key lookup. The reserved set is whatever `helm.annotations.vault` and `helm.otel.annotations` happen to emit (`_spec_cronjob.tpl:51`), dumped at the head SHA via a debug template as exactly ten keys: `mutate-probes`, `run-as-group`, `run-as-non-root`, `run-as-user`, `vault-addr`, `vault-agent-env-variables`, `vault-auth-method`, `vault-env-daemon`, `vault-path`, `vault-role`.

Line 682 says "all `vault.security.banzaicloud.io/*` (pod template)". That is not what the code does, and every other key in that namespace passes through onto the pod template — **including on the CronJob surface this PR fixes**. Rendered at the head SHA:

```yaml
vault.security.banzaicloud.io/mutate: skip
vault.security.banzaicloud.io/vault-namespace: root
vault.security.banzaicloud.io/vault-serviceaccount: vault-admin
vault.security.banzaicloud.io/vault-skip-verify: "true"
```

From the [bank-vaults annotation reference](https://bank-vaults.dev/docs/mutating-webhook/annotations/), quoting the page verbatim:

| annotation | default | vendor text |
|---|---|---|
| `vault-skip-verify` | `"false"` | "Same as VAULT_SKIP_VERIFY" |
| `mutate` | `""` | "possible values: `skip` which prevents it" |

`vault-namespace` and `vault-serviceaccount` are also documented; both are unreserved.

**These are two different risk classes and should not be conflated.** `vault-skip-verify` is a *confidentiality* control — disabling TLS verification on the connection to Vault enables MITM of secret retrieval. `mutate: skip` is an *availability / fail-closed* outcome — the webhook does not mutate the pod, `vault:` env references are never resolved, and the workload starts with literal placeholder strings and typically crashes. Only the first is a security-posture weakening.

*(Separately, and **not** vendor-documented: `vault.security.banzaicloud.io/psp-allow-privilege-escalation` also renders through. It does not appear in the current bank-vaults reference, so it is recorded as observed output only, with no claim about its effect.)*

**Three further documentation inaccuracies on these same lines**, each verified by render at the head SHA rather than by reading the templates:

1. `instrumentation.opentelemetry.io/*` is wrong — only `container-names` and `inject-<resolved-language>` are emitted. With `global.language: nodejs`, a user's `instrumentation.opentelemetry.io/inject-java` renders through alongside the chart's `inject-nodejs`.
2. `manualOtelConfig: true` is a **third** condition on the otel keys (`_otel.tpl:19-25` gates `_otel.tpl:37-45`). Verified: with `manualOtelConfig: true` and autoinstrumentation enabled, a user's `sidecar.opentelemetry.io/inject: "false"` renders onto the pod template.
3. `vault-env-daemon` is itself conditional on `.Values.secrets` existing (`_vault.tpl:12`), and `vault-env-from-path` only appears when `secrets.bulk.path` is set — so "the ten keys" is only true under those conditions.

**On the fix:** a blanket prefix-match was considered and **withdrawn**. Reserving the whole `vault.security.banzaicloud.io/*` namespace would silently *drop* keys the chart never sets, including `vault-namespace` and `vault-serviceaccount` which the vendor documents as legitimate per-workload knobs — trading silent-override for silent-drop on a wider surface, with no error and no escape hatch. The proportionate change is a small explicit deny-list of the security-relevant keys the chart does not set (`vault-skip-verify`, `mutate`, `vault-tls-secret`), which closes the MITM vector without foreclosing the documented knobs.

The suggestion below fixes the documentation to match today's behaviour, which is the minimum this PR should ship.

```suggestion
## and it will not render twice. Keys are matched EXACTLY, not by prefix. The reserved set is only
## what the chart actually renders, and it is context-dependent:
##   - pod template, from `helm.annotations.vault`: vault-addr, vault-auth-method, vault-path,
##     vault-role, vault-agent-env-variables, mutate-probes, run-as-non-root, run-as-user,
##     run-as-group; plus vault-env-daemon only when `secrets` is set, and vault-env-from-path only
##     when `secrets.bulk.path` is set.
##   - pod template, from `helm.otel.annotations`: `sidecar.opentelemetry.io/inject`,
##     `instrumentation.opentelemetry.io/container-names` and
##     `instrumentation.opentelemetry.io/inject-<resolved-language>` (that one language only) —
##     and ONLY when `manualOtelConfig` is not true AND `disableOtelAutoinstrumentation: false`
##     AND the language is nodejs/java/python.
##   - CronJob metadata: `argocd.argoproj.io/sync-options`.
##
## WARNING: any other key IS applied to the pod, including other `vault.security.banzaicloud.io/*`
## keys the chart does not set — notably `vault-skip-verify` (disables TLS verification to Vault)
## and `mutate: skip` (disables secret injection entirely).
```

charts/mycarrier-helm/values.yaml
```diff
@@ -588,8 +588,20 @@ applications: {}
...
+## **Security Context (opt-in)**: securityContext.readOnlyRootFilesystem (default false; sets the
+## container's readOnlyRootFilesystem), securityContext.fsGroup (unset by default; sets the pod-level
+## fsGroup; must be >= 1 — fsGroup 0 is the root group and is rejected by the values schema),
+## securityContext.seccompProfile (unset by default; sets the pod-level seccompProfile, e.g.
+## {type: RuntimeDefault}). All three are opt-in per job item — omitting securityContext entirely
+## preserves the chart's existing default rendering.
+##
+## IMPORTANT: the cluster's kyverno `restrict-seccomp` policy is Enforce and only allows
+## seccompProfile.type `RuntimeDefault` (or `runtime/default`) — `Localhost` renders fine from this
+## chart but is hard-rejected at admission. Stick to `RuntimeDefault`.
```
smunukutla-mycarrier: **🟠 important · security · Security-Other**

### This prose asserts an admission control that does not exist — `restrict-seccomp` is Audit, and its pattern rejects `RuntimeDefault`

All three assertions in these added lines were checked against `MyCarrier-DevOps/kyverno-policies` at `main`. All three are false.

**1. "is Enforce"** — `restrictions/restrict-seccomp.yaml` has `validationFailureAction: Audit`. The repo README confirms this is the deliberate lifecycle: *"Every policy in this repo starts life as `validationFailureAction: Audit` (report-only — violations show up in `PolicyReport` objects but nothing is blocked)"*, and that `kyverno-policies@main` is rendered onto all three app clusters (cluster-dev, cluster-data, cluster-production-csp). A repo-wide search finds no `Enforce` anywhere and no `validationFailureActionOverrides` on this policy. **Nothing is blocked at admission today.**

**2. "only allows seccompProfile.type `RuntimeDefault` (or `runtime/default`)"** — the policy's `validate.pattern` is `=(type): "runtime/default"` on all three of `spec.securityContext`, `initContainers` and `containers`. Kyverno patterns match literally. `RuntimeDefault` does not match `runtime/default`, so the relationship is **the inverse of what this comment states**: the policy accepts only `runtime/default`, which is *not* a legal value for the Kubernetes `SeccompProfile.type` field (valid values are `RuntimeDefault`, `Localhost`, `Unconfined` — `runtime/default` is the legacy seccomp *annotation* value, removed in k8s 1.25).

Following this PR's own advice to "Stick to `RuntimeDefault`" therefore **records a PolicyReport violation** rather than passing the check.

**3. "`Localhost` renders fine from this chart but is hard-rejected at admission"** — nothing is hard-rejected, because the action is Audit. Separately, PSS `restricted` *permits* `Localhost` (it forbids only `Unconfined`), so the security framing is wrong in both directions.

**Why this matters beyond accuracy:** operator-facing documentation that presents an audit-only control as an enforcement gate is actively harmful. It tells operators a class of misconfiguration will be caught for them at admission when it will not, and it discourages the local testing that would catch it. That is the same failure mode as the rest of this PR — a security expectation that is documented but not actually wired up.

**Note:** the identical block also exists at `values.yaml:670-672` in the cronjobs section (pre-existing, from #110, so not anchorable here). It needs the same correction — otherwise the chart contradicts itself.

```suggestion
## NOTE: the cluster's kyverno `restrict-seccomp` policy is `validationFailureAction: Audit`
## (report-only — violations appear in PolicyReport objects, nothing is blocked at admission), and
## its pattern matches the literal `runtime/default`, which is not a legal value for the Kubernetes
## seccompProfile.type field. Use `RuntimeDefault` (the correct API value; PSS Restricted forbids
## only `Unconfined`) and expect a PolicyReport entry until that policy is corrected.
```

charts/mycarrier-helm/values.schema.json
```diff
@@ -1725,6 +1726,26 @@
...
+              "fsGroup": {
+                "type": "integer",
+                "description": "Pod-level fsGroup (opt-in; unset by default; must be >= 1 — fsGroup 0 is the root group and is rejected)",
+                "minimum": 1
+              },
+              "seccompProfile": {
+                "type": "object",
+                "description": "Pod-level seccompProfile (opt-in; unset by default)"
+              }
+            }
```
smunukutla-mycarrier: **💡 suggestion · security · Security-Insecure-Default**

### `jobs[].securityContext` accepts unknown keys and an empty `seccompProfile` — hardening silently discarded

*This block merges two findings (unknown-key acceptance and `seccompProfile` under-specification) because they target the same object and GitHub cannot apply two overlapping suggestions.*

The new block (`values.schema.json:1730-1748`) sets no `additionalProperties` constraint, and `seccompProfile` is a bare `{"type":"object"}` with no `properties` and no `required`. **Three ways a hardening value validates and then vanishes**, all reproduced at the head SHA:

1. `readOnlyRootFileSystem: true` (capital S) validates; the container renders `readOnlyRootFilesystem: false`. The operator asked for a read-only root filesystem and did not get one.
2. `seccompProfile: {}` validates; `{{- with $sc.seccompProfile }}` (`_podsecurity.tpl:31`) is falsey on an empty map, so the key is dropped.
3. `seccompProfile: {typo: RuntimeDefault}` validates **and renders into the manifest**, which the API server then rejects at apply time — the failure lands far from the mistake.

This is the same silent-drop class the PR exists to eliminate.

**Blast radius**, since the schema is consumed unversioned: `charts/mc-environment/values.schema.json:335` remote-`$ref`s `https://raw.githubusercontent.com/MyCarrier-DevOps/helm-charts/refs/heads/main/charts/mycarrier-helm/values.schema.json#/properties/jobs` — an **unpinned ref to `main`**, so any constraint added here reaches every mc-environment consumer **on merge**, with no chart-version gate. Priced: `gh search code --owner MyCarrier-DevOps` finds no mycarrier-helm consumer setting `jobs[].securityContext` anywhere, so the live blast radius today is zero. The structural exposure is real regardless, and pinning that `$ref` to a tag is worth doing independently of this PR. If you'd rather not take the constraint now, stage it — declare the missing keys first, add `additionalProperties: false` in a later announced change.

**Validated against the chart before proposing:** `helm unittest` stays green at 59 suites / 607 tests / 6 snapshots; variants 1–3 now fail at schema time with precise messages (`additional properties 'readOnlyRootFileSystem' ... not allowed`, `missing property 'type'`, `additional properties 'typo' not allowed`); and `seccompProfile: {type: RuntimeDefault}` still renders correctly. The replacement below was re-checked as valid JSON in place.

The enum is recommended on **Kubernetes API grounds** (`SeccompProfile.type` is `RuntimeDefault` | `Localhost` | `Unconfined`) — *not* on kyverno grounds; see the separate comment on `values.yaml:601-603`, where that policy turns out to be Audit-only and to require a string that is not a legal API value. A team targeting PSS `restricted` could narrow the enum to `["RuntimeDefault","Localhost"]`; all three legal API values are kept here so the schema does not reject a manifest Kubernetes accepts.

The same gap exists on the cronjobs block (`:2001-2020`) and the applications block (`:600-617`); the latter needs `addCapabilities` declared first, since `_podsecurity.tpl:52` reads it but no schema declares it.

```suggestion
              "seccompProfile": {
                "type": "object",
                "description": "Pod-level seccompProfile (opt-in; unset by default)",
                "properties": {
                  "type": {
                    "type": "string",
                    "enum": ["RuntimeDefault", "Localhost", "Unconfined"],
                    "description": "Required by the Kubernetes SeccompProfile API"
                  },
                  "localhostProfile": { "type": "string" }
                },
                "required": ["type"],
                "additionalProperties": false
              }
            },
            "additionalProperties": false
```

charts/mycarrier-helm/templates/_annotations.tpl
```diff
@@ -0,0 +1,27 @@
...
+
+{{/*
+The chart-managed CronJob *metadata* annotations (as opposed to the pod-template annotations
+covered by helm.annotations.vault / helm.otel.annotations). Kept as its own define so
+cronjob.yaml can derive its reserved-key set from the exact same source it renders from,
+rather than hardcoding the key twice and risking drift.
+*/}}
+{{- define "helm.annotations.cronjobMetadata" -}}
+argocd.argoproj.io/sync-options: Prune=true
+{{- end -}}
```
smunukutla-mycarrier, line 39: **💡 suggestion · technical-reviewer · Contract**

### Job metadata needs the same treatment as CronJob metadata, and the remedy does not generalise as one define

Same defect class as the comment on `_annotations.tpl:6-17`, which already lists `jobs.yaml:24-26` among its surfaces. Raised separately only for two facts that comment does not carry.

**1. The remedy is NOT one define.** `helm.annotations.cronjobMetadata` works because the CronJob metadata block is a single static line. The Job path has three different chart-managed sync-options values:

```
cronjob.yaml:23        argocd.argoproj.io/sync-options: Prune=true
jobs.yaml:23           argocd.argoproj.io/sync-options: Force=true,Replace=true
_spec_job.tpl:21       argocd.argoproj.io/sync-options: Prune=true,Replace=true
```

plus `_spec_job.tpl:22-32`, where `argocd.argoproj.io/hook` and `.../sync-wave` are emitted conditionally on `.job.timing`. A "derive the reserved set from the same source you render from" extraction needs **three** defines, and the Job pod-template one is conditional. Anyone copying the `cronjobMetadata` pattern verbatim will get it wrong.

**2. The Trivy document-skip IS reproducible on the Job surface** — run rather than asserted. Trivy 0.72.0, isolated single-document Job renders at the head SHA:

```
jobs[0].annotations."argocd.argoproj.io/sync-options": Prune=false  -> Results: 0
jobs[0].annotations.team: platform (no collision)                   -> Results: 1, 6 misconfigs
  KSV-0014 HIGH "Root file system is not read-only", KSV-0020, KSV-0021,
  KSV-0030, KSV-0104, KSV-0125
```

So the duplicate key silently suppresses a HIGH finding on the Job exactly as it does on the Deployment.

Downgraded from *important*: this is pre-existing on a resource this PR does not touch, and the same class is called non-blocking above.

charts/mycarrier-helm/templates/_spec_cronjob.tpl
```diff
@@ -48,8 +48,10 @@ jobTemplate:
         annotations:
           {{ include "helm.annotations.vault" . | indent 10 | trim }}
           {{ include "helm.otel.annotations" . | indent 10 | trim }}
-          {{- with .cronjob.annotations }}
-          {{ toYaml . | indent 10 | trim }}
+          {{- $reserved := merge (dict) (include "helm.annotations.vault" $ | fromYaml | default dict) (include "helm.otel.annotations" $ | fromYaml | default dict) }}
+          {{- $extra := include "helm.annotations.userExtra" (dict "reserved" $reserved "user" $.cronjob.annotations) }}
+          {{- if $extra }}
+          {{ $extra | indent 10 | trim }}
```
smunukutla-mycarrier: **💡 suggestion · technical-reviewer · SOLID-OCP**

### Pod-template reserved set duplicates the render list by hand; the test suite does not hold the coupling

**Correction to an earlier draft of this review:** the suite does *not* stay green when a helper is added. The drift experiment, run properly — adding a fourth chart-managed helper (`mycarrier.io/drift-demo`) to the cronjob pod template after line 50, leaving line 51 alone:

```
helm lint     -> clean
helm unittest -> Tests: 2 failed, 605 passed, 607 total   (helm-unittest v1.1.0, the CI pin)
```

So the suite does bite. That is a genuine strength of these tests.

**The finding survives on what happens next.** The two failures are an exact-map `equal` (`cronjob_test.yaml:822`) and a snapshot — both of which say *"the rendered annotation map changed"*, not *"the reserved set is stale"*. The natural developer response is to add the new key to the expected map and run `helm unittest -u`. Doing exactly that:

```
helm unittest -> Tests: 607 passed, 607 total / Snapshot: 6 passed, 6 total
```

and the bug is live. Rendering a CronJob whose user annotations set the new key:

```
64:            mycarrier.io/drift-demo: "yes"
65:            mycarrier.io/drift-demo: USER-WINS
```

A duplicate key in the pod template — precisely the defect this PR exists to remove — with a fully green suite. The tests detect that output *changed*; nothing detects that line 51 no longer matches lines 49-50.

Root cause: line 51 re-enumerates by hand the helper list rendered on lines 49-50, and nothing couples them. Contrast `helm.annotations.cronjobMetadata` (`_annotations.tpl:19-23`), extracted specifically to avoid "hardcoding the key twice and risking drift" — the pod-template path, the more complex of the two, did not get that treatment.

*(The `.` on line 49 vs `$` on line 51 is verified inert today — `helm.annotations.vault` reads only `.Values`, and `helm.getLanguage` falls through to `.Values.global.language` on both since neither context carries an `application` key. Noted only as a reason to remove the divergence while refactoring, not as a present defect.)*

**Remedy, unified with the `fromYaml` comment on `cronjob.yaml:24-25` so the two do not conflict:** build the chart-managed annotations as a **dict** once, above line 48, `toYaml` it for the render on lines 49-50, and use its keys directly for the reserved set. That removes the hand-maintained list *and* the `fromYaml` round-trip in one change. Not a localized edit (it must be inserted before unchanged lines 49-50), so no suggestion block.

charts/mycarrier-helm/templates/_annotations.tpl
```diff
@@ -0,0 +1,27 @@
+{{/*
+Emit the user-supplied annotations that do NOT collide with chart-managed keys.
+Chart-managed keys always win: a colliding user key is dropped rather than emitted
+as a duplicate YAML key. Returns "" when nothing survives, so callers can guard with `if`.
+*/}}
```
smunukutla-mycarrier, line 5: **💡 suggestion · technical-reviewer · Contract**

### CronJob sync-options is now hard-locked to a single non-documented token with no override field

**Three claims from an earlier draft are withdrawn**, and they should not be weighed:

- *"CronJob `spec.schedule`/`jobTemplate` changes can require `Replace=true`"* — wrong. From `kubernetes/kubernetes` `pkg/apis/batch/validation/validation.go`, `ValidateJobSpecUpdate` calls `validatePodTemplateUpdate` and enforces `Selector` immutability — **Job** pod templates are immutable, which is why `jobs.yaml:23` and `_spec_job.tpl:21` carry `Replace=true`. `ValidateCronJobUpdate` performs no old-vs-new immutability comparison on `spec.jobTemplate` or `spec.schedule`; CronJob changes update in place.
- *"Any consumer who set this is currently getting their value applied"* — `gh search code --owner MyCarrier-DevOps` found no mycarrier-helm values file setting `sync-options` inside a `cronjobs:` block, corroborating the PR body's blast-radius claim.
- *"Let the user value win for this key"* — that would reintroduce the duplicate-key/Trivy-skip condition reproduced elsewhere in this review. Only an explicit field works.

**What remains, verified this round:** `argocd.argoproj.io/sync-options` is a comma-separated multi-value field. The ArgoCD Sync Options page documents 16 values — `Validate=false`, `SkipDryRunOnMissingResource=true`, `Delete=false`, `ApplyOutOfSyncOnly=true`, `PrunePropagationPolicy=...`, `PruneLast=true`, `Replace=true`, `Force=true`, `ServerSideApply=true`, `ClientSideApplyMigration=false`, `FailOnSharedResource=true`, `RespectIgnoreDifferences=true`, `CreateNamespace=true`, `Prune=false`, `Prune=confirm` — and **`Prune=true` is not among them**; only `Prune=false` and `Prune=confirm` are documented.

So this PR makes a token that is not a documented ArgoCD option the *sole permitted value* of a field designed to hold several, and closes the only route (however accidental) by which a consumer could add `ServerSideApply=true`, `PruneLast=true` or `Validate=false` to a CronJob.

Forward-looking API-surface gap, not a live regression. Consider adding `cronjobs[].syncOptions` (appended to the chart-managed value rather than replacing it), and separately confirm whether `Prune=true` is intended at all. At minimum the `values.yaml` prose should say the field is now closed, since it currently only says colliding keys are dropped.

charts/mycarrier-helm/values.schema.json
```diff
@@ -606,7 +606,8 @@
               },
               "fsGroup": {
                 "type": "integer",
-                "description": "Pod-level fsGroup (opt-in; unset by default)"
+                "description": "Pod-level fsGroup (opt-in; unset by default; must be >= 1 — fsGroup 0 is the root group and is rejected)",
+                "minimum": 1
```
smunukutla-mycarrier: **💡 suggestion · technical-reviewer · Logic**

### The schema bound is right, but the falsey-`with` guard that made it necessary is still there

**Two claims from an earlier draft are withdrawn:**

1. *"Banning `fsGroup: 0` is a policy choice the chart is now imposing"* — wrong; the org already declares it. `MyCarrier-DevOps/kyverno-policies` `restrictions/require-non-root-groups.yaml`, rule `check-fsGroup`: *"spec.securityContext.fsGroup must be empty or greater than zero"*, category "Pod Security Standards (Restricted)". `"minimum": 1` restates the platform's own bound exactly. (For balance: that policy is `validationFailureAction: Audit`, so nothing is blocked at admission today — it reports into `PolicyReport` objects.)
2. *"A values file that deploys today hard-fails after upgrade"* — named no file, and an org-wide search found no mycarrier-helm consumer setting `securityContext.fsGroup` anywhere. The render-time failure changes no deployed output, because `fsGroup: 0` was already a silent no-op.

Consequently the previously proposed remedy is also withdrawn: `{{- if not (kindIs "invalid" $sc.fsGroup) }}` **alone** would make `fsGroup: 0` render as `fsGroup: 0`, producing a manifest that `check-fsGroup` flags. That is worse than the current behaviour, not better.

**What survives:** the schema constraint fixes one value of one field on three surfaces; the defective idiom is untouched. `_podsecurity.tpl:28` guards a numeric with `{{- with }}`, which is a falsey test, not a presence test. Evidence that this is a class and not an instance: the very next block, `_podsecurity.tpl:31`, has the identical untreated defect — `seccompProfile: {}` validates and then silently vanishes (see the comment on `values.schema.json:1744`). Any future falsey-but-legal value added under this idiom repeats the bug, and if `minimum` is ever relaxed the original defect returns unannounced.

*(The PSS point stands and was confirmed from `kubernetes/website`: `pod-security-standards.md` contains zero occurrences of `fsGroup`, `runAsGroup` or `supplementalGroups`. The bound is org policy, not an upstream standard — worth stating in the `description` so the next reader knows where it came from. A parallel check of the CIS Kubernetes Benchmark v1.10 §5.2 via `kube-bench`'s reference implementation also returned zero GID-related controls.)*

**Recommended:** keep `"minimum": 1` **and** change the guard to presence-based (`{{- if hasKey $sc "fsGroup" }}`). Then the schema is the single place the legal range is expressed, and the template stops having an opinion about which values are "empty". `_podsecurity.tpl:28` is not in this diff, so no suggestion block.

charts/mycarrier-helm/values.yaml
```diff
@@ -109,7 +109,7 @@ deployment: deployment
...
 ## **Resources**: resources.requests.cpu/memory, resources.limits.cpu/memory
 ## **Health Probes**: probes.enableLiveness, probes.enableReadiness, probes.enableStartup (all default true; set to false to disable), custom probe configurations
-## **Security Context**: securityContext.readOnlyRootFilesystem (default false), securityContext.addCapabilities (default []; all capabilities are dropped by default, list specific capabilities here to add them back), securityContext.fsGroup (opt-in pod-level fsGroup; previously ignored/silently dropped, now honored), securityContext.seccompProfile (opt-in pod-level seccompProfile; previously ignored/silently dropped, now honored)
+## **Security Context**: securityContext.readOnlyRootFilesystem (default false), securityContext.addCapabilities (default []; all capabilities are dropped by default, list specific capabilities here to add them back), securityContext.fsGroup (opt-in pod-level fsGroup; previously ignored/silently dropped, now honored; must be >= 1 — fsGroup 0 is the root group and is rejected by the values schema), securityContext.seccompProfile (opt-in pod-level seccompProfile; previously ignored/silently dropped, now honored)
```
smunukutla-mycarrier: **💡 suggestion · technical-reviewer · Convention**

### Three behaviour changes with no changelog entry — and the schema changes bypass version pinning entirely

**Withdrawn from an earlier draft:** *"a patch release gives operators no signal to review"*. Consumers pin exactly — `charts/mc-environment/templates/appset.yaml:64` is `targetRevision: "{{ $.Values.mycarrierChartVersion | default "3.4.0" }}"`, set by `workflow-core` via `--set-string`. A 3.5.8 tag auto-upgrades nobody, and the signal does arrive when the pin moves. So the minor-bump ask is not the right ask. (Corrected citation: `manualChartVersionDetected` is at `bump_chart_version.js` line **59**, not 58.)

**The three behaviour changes are still real** and each was reproduced:

1. Annotation precedence reversal on CronJob metadata and pod template — base yields the user's `Prune=false`, head yields the chart's `Prune=true`.
2. `fsGroup: 0` renders on base, hard-fails at head with `at '/cronjobs/0/securityContext/fsGroup': minimum: got 0, want 1`.
3. `jobs[].securityContext` inert on base, active at head; `jobs.items` had no `additionalProperties` at base, so the key did validate before.

**The sharper point, which inverts the remedy:** the schema changes are **not gated by any chart version**. `charts/mc-environment/values.schema.json:332,335,338` remote-`$ref` `.../refs/heads/main/charts/mycarrier-helm/values.schema.json#/properties/{applications,jobs,cronjobs}` — an unpinned floating ref to `main`.

This was verified end to end, not inferred. Under Helm **3.18.6** (the version `unit-tests.yaml:70-73` pins) and 3.16.2, 3.14.4 and 4.2.2, Helm *does* fetch and enforce remote `$ref`s. Rendering `charts/mc-environment` with a cronjobs array carrying an invalid `securityContext.fsGroup`:

```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
mc-environment:
- at '/environments/0/cronjobs': 'oneOf' failed, none matched
  - at '/environments/0/cronjobs/0/securityContext/fsGroup': got string, want integer
```

That constraint exists **only** in `mycarrier-helm/values.schema.json` on `refs/heads/main`. So `"minimum": 1` reaches every mc-environment consumer the instant this merges, regardless of `mycarrierChartVersion`. Neither 3.5.8 nor 3.6.0 gates it, and the exact-version pinning that defuses the template changes does nothing here.

**Ask:** (a) a changelog / release-note entry covering the three changes, and (b) treat the floating `$ref` as the real versioning defect — schema edits in this repo are effectively released to mc-environment on merge. That is also why the `additionalProperties: false` proposal on `values.schema.json:1744` may warrant staging.

*Anchor note: this concerns `Chart.yaml` and `mc-environment`, neither in the diff. Line 112 is the line this PR modified to announce two of the three changes ("previously ignored/silently dropped, now honored… rejected by the values schema"), so it is the closest honest anchor.*

charts/mycarrier-helm/templates/cronjob.yaml
```diff
@@ -20,9 +20,11 @@ metadata:
     {{ toYaml .labels | indent 4 | trim }}
     {{- end }}
   annotations:
-    argocd.argoproj.io/sync-options: Prune=true
-  {{- with .annotations }}
-    {{ toYaml . | indent 4 | trim }}
+    {{ include "helm.annotations.cronjobMetadata" . | indent 4 | trim }}
+  {{- $reserved := include "helm.annotations.cronjobMetadata" . | fromYaml | default dict }}
+  {{- $extra := include "helm.annotations.userExtra" (dict "reserved" $reserved "user" .annotations) }}
```
smunukutla-mycarrier: **💡 suggestion · technical-reviewer · Error-Handling**

### `fromYaml | default dict` is dead code on both branches; the round-trip is avoidable entirely

Helm's `fromYaml` returns `map[string]interface{}{"Error": "<parse error>"}` on failure, so `| default dict` can never fire on the error path (non-empty map); on success with empty input it returns `{}`, where `default dict` yields an equally empty dict. **Dead on both branches.** Same pattern at `_spec_cronjob.tpl:51`. Verified under Helm v4.2.2:

```
fromYaml "a: 1\n  b: 2\n"
  -> {"Error":"error converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context"}
  hasKey $r "Error" == true ; default dict never fires
```

*Correction to an earlier draft: "every `hasKey` lookup returns false" is not quite right — `hasKey $reserved "Error"` returns true, so a user annotation literally keyed `Error` would still be dropped. Immaterial, but the finding should be accurate.*

**Reachability, stated plainly:** no trigger was found, by any of the three reviewers. The string fed to `fromYaml` and the string rendered into the manifest come from the same `include`, so any input that breaks the parse also breaks the manifest; the only divergence vector is indentation (`indent 10 | trim` vs raw) and no one could construct one. This is a defense-in-depth guard against a latent failure mode, not a live bug — which is why it is a suggestion rather than something to block on.

**Remedy, unified with the comment on `_spec_cronjob.tpl:51-54` so the two do not give contradictory instructions:** do not guard the parse — *remove* it. Build the chart-managed annotations as a dict, `toYaml` it once for the render, and take the reserved keys from the dict directly. No round-trip, no `Error` branch, no hand-maintained helper list, and the render and the reserved set become the same object by construction.

If the round-trip is kept for any reason, then add `{{- if hasKey $reserved "Error" }}{{ fail ... }}{{ end }}`; that fires only on chart-authored content and does not conflict with the deliberate "never fail on a user collision" policy.

charts/mycarrier-helm/templates/_podsecurity.tpl
```diff
@@ -12,6 +12,9 @@ below are no-ops and rendered output is unchanged).
 {{- if and .cronjob .cronjob.securityContext -}}
 {{- $sc = .cronjob.securityContext -}}
 {{- end -}}
+{{- if and .job .job.securityContext -}}
+{{- $sc = .job.securityContext -}}
+{{- end -}}
```
smunukutla-mycarrier: **💡 suggestion · technical-reviewer · SOLID-OCP**

### Three independent `if`s encode an undocumented last-write-wins precedence; the chart already uses the safer idiom elsewhere

The fix itself is correct and the tests are not tautological — stripping the `.job` branch makes `job_test.yaml` fail exactly 3 assertions (`Tests: 3 failed, 34 passed, 37 total`), independently confirmed.

**Two corrections to an earlier draft.** The call-site count is **15**, not 12: `_spec_rollout.tpl:87,111,170`; `_spec_job.tpl:36,94`; `_spec_statefulset.tpl:39,74,132`; `_spec_triggertestengine.tpl:41,163`; `_spec_deployment.tpl:77,101,161`; `_spec_cronjob.tpl:57,115`. All 15 were re-walked and the conclusion holds — no context carries two workload-type keys, so the precedence is unreachable today.

And `_labels.tpl` was mischaracterised. It uses `{{- else if .job -}}` (`:49`) and `{{- else if .cronjob -}}` (`:59`) — an **exclusive** chain. It is the *counter-example*, not the precedent: the chart already knows the safer idiom, and `_podsecurity.tpl:9-17` departs from it with three independent `if`s, which is what creates the silent `application → cronjob → job` override ordering that nothing documents or tests.

That makes the remedy much smaller than a caller-passes-the-dict refactor: **convert lines 12 and 15 to `else if`**, matching `_labels.tpl`. Exclusivity then becomes structural, a context carrying two keys fails loudly instead of silently picking the last branch, and the diff is two words. Not expressible as a suggestion block because it also requires editing line 12, which is not in this diff.

charts/mycarrier-helm/values.schema.json
smunukutla-mycarrier: **💡 suggestion · technical-reviewer · DRY**

### Third verbatim copy of the securityContext block; the copies have already diverged

This PR adds a third literal copy of the same `securityContext` object (applications `:601-616`, jobs `:1730-1748`, cronjobs `:2004-2019`). Zero `$ref`/`$defs` in the file, so nothing keeps them in sync — and they have already drifted:

- applications' `readOnlyRootFilesystem` has no `description`; jobs' and cronjobs' do.
- `addCapabilities` is documented at `values.yaml:112` and read at `_podsecurity.tpl:52`, but is absent from all three blocks.
- None sets `additionalProperties: false`. See the comment on `values.schema.json:1744`, which proposes exactly that on this block — these two agree; treat `additionalProperties` as one decision for all three copies rather than one.

**The objection that `$defs` would break `mc-environment` does not hold** — it was tested rather than reasoned about. `mc-environment` remote-`$ref`s *sub-paths* of this file (e.g. `#/properties/applications/patternProperties/^[a-z0-9_-]+$/properties/networking`, `:344-350`). The equivalent was built: a schema served over HTTP whose `properties.jobs` internally refs `"#/$defs/sc"`, and a consumer chart whose own schema refs `"http://.../a.schema.json#/properties/jobs"`. Result:

```
fsGroup: 5 -> renders
fsGroup: 0 -> "at '/jobs/0/securityContext/fsGroup': minimum: got 0, want 1"
```

The local pointer resolves against the remote document's base URI, as the spec requires; it does not dangle. This was independently confirmed a second way (nested local refs two levels deep inside a fetched document resolve correctly under both Helm 3.18.6 and 4.2.2).

Extracting `$defs.workloadSecurityContext` and `$ref`-ing it from all three sites removes the drift vector, lets `addCapabilities` be declared once, and makes `additionalProperties` a single decision. Repo guidance (`.github/copilot-instructions.md:22`, "DRY Principle") applies.

**Staging caveat:** because `mc-environment`'s `$ref` floats on `refs/heads/main`, any *tightening* here lands on those consumers on merge with no version gate. Declare missing keys first; add `additionalProperties: false` in a later, announced change.

charts/mycarrier-helm/values.yaml
```diff
@@ -588,8 +588,20 @@ applications: {}
 ##
 ## Each job supports: name, timing (pre-deploy/post-deploy), order, ttlSecondsAfterFinished,
 ## activeDeadlineSeconds, backoffLimit, labels, annotations, image (registry/repository/tag),
-## imagePullPolicy, command, args, env, resources (requests/limits for cpu/memory)
-## 
+## imagePullPolicy, command, args, env, resources (requests/limits for cpu/memory), volumes,
+## securityContext (opt-in hardened security context, see below)
```
smunukutla-mycarrier, line 624: **💡 suggestion · technical-reviewer · Convention**

### Jobs cannot set pod annotations at all, and the hazard is worse for Jobs than for CronJobs

The cronjobs section now carries ~17 lines (`values.yaml:674-690`) explaining that annotations land on both the resource and the pod template, and specifically that this is how you set `sidecar.istio.io/inject: "false"` to stop a never-exiting Istio sidecar hanging the Job. The jobs section, edited in this same hunk, still lists `annotations` bare.

`_spec_job.tpl:20-34` never references `.job.annotations`. Rendered and verified: `jobs[0].annotations` appear on `metadata.annotations` only, never on `spec.template.metadata.annotations`.

**"No workaround available" was challenged; here is the substantiation.** `grep -n "istioDisabled\|annotations.istio" _spec_job.tpl jobs.yaml` returns nothing — there is no istio reference anywhere in the Job path. `jobs[]` is a top-level array, not nested under an application, so `applications[].istioDisabled` (read at `_helpers.tpl:203` via `.application`) **cannot reach a `jobs[]` item**; there is no application context to carry it. A namespace-level `istio-injection` label is not a per-Job control — it would disable injection for every workload in the namespace. Neither candidate workaround applies.

**Recommendation inverted from an earlier draft**, which treated Job and CronJob as equivalent. They are not, and the asymmetry favours Jobs: Jobs from this chart are ArgoCD sync hooks (`argocd.argoproj.io/hook: PreSync|PostSync`, `hook-delete-policy: BeforeHookCreation`, `_spec_job.tpl:21-32`). **A hung PreSync hook blocks the entire application sync**; a hung CronJob-spawned pod blocks one tick. The primary recommendation is therefore to *close* the gap — render `.job.annotations` onto the Job pod template with the same `userExtra` filtering — with the doc note below as the fallback if that is deferred.

`_spec_triggertestengine.tpl` renders a fourth Job-shaped hook workload (`argocd.argoproj.io/hook: PostSync`, `triggertestengine.yaml:21`) with the same gap and no annotation input at all.

```suggestion
## imagePullPolicy, command, args, env, resources (requests/limits for cpu/memory), volumes,
## securityContext (opt-in hardened security context, see below)
##
## NOTE: unlike `cronjobs[].annotations`, `jobs[].annotations` render ONLY onto the Job resource
## metadata — the Job pod template (`_spec_job.tpl`) never applies them. There is currently no way
## to set a per-Job pod annotation from this chart, including `sidecar.istio.io/inject: "false"`.
## This matters more for Jobs than for CronJobs: chart Jobs are ArgoCD PreSync/PostSync hooks, so a
## never-exiting Istio sidecar blocks the whole application sync rather than one scheduled run.
```

charts/mycarrier-helm/values.yaml
```diff
@@ -643,18 +662,32 @@ jobs: []
...
 ##
-## `annotations` now also render onto the CronJob pod template (in addition to the CronJob resource
-## itself), so it can be used to set per-cronjob pod annotations — e.g. `sidecar.istio.io/inject: "false"`
-## to opt a CronJob's pods out of Istio sidecar injection in istio-injection=enabled namespaces (a
-## never-exiting sidecar container otherwise hangs the Job).
+## `annotations` render onto BOTH the CronJob resource metadata and the CronJob pod template, so it
+## can be used to set per-cronjob pod annotations — e.g. `sidecar.istio.io/inject: "false"` to opt a
+## CronJob's pods out of Istio sidecar injection in istio-injection=enabled namespaces (a never-exiting
+## sidecar container otherwise hangs the Job). `sidecar.istio.io/inject` is not chart-managed and
+## passes through unaffected.
```
smunukutla-mycarrier: **💡 suggestion · technical-reviewer · Convention**

### New behaviour documented as free-form prose, which the README generator cannot pick up

**Two corrections to an earlier draft**, so they are not weighed: the CronJob Configuration Options table (`README.md:1035-1057`) lists **21** fields, not 18. And the helper-function limb is withdrawn entirely — that section opens *"These are the main helper functions in the templates"* and shows three examples out of **141** `define`s; `helm.annotations.vault` and `helm.podSecurityContext` appear zero times. The absence of `helm.annotations.userExtra` is the norm, not a gap.

**Reframed around the repo's own mechanism, which is the actual point.** `.github/workflows/readme-gen.yaml` triggers on `pull_request_target` for `charts/*/values.yaml`, runs `bitnami/readme-generator-for-helm@2.7.2` against `values.yaml` + `README.md`, and commits the result back to the PR branch. It regenerates the `## Parameters` section (`README.md:1277`) from `## @param` annotations.

This PR adds 30+ lines of behavioural documentation to `values.yaml` as free-form `##` prose and adds **zero** `## @param` lines (verified: `git diff | grep '^+' | grep -c '@param'` = 0). So the generator produces no README diff, and the hand-written sections stay stale:

- the CronJob table has no `annotations`, `labels` or `securityContext` row;
- the "Advanced CronJob with All Features" example (`README.md:946-951`) shows `annotations:` with no hint that chart-managed keys now silently win;
- `grep -c securityContext README.md` returns **0** — the feature added in #110 and extended here is absent from the README entirely.

Repo guidance (`.github/copilot-instructions.md:36-37`): *"Update README.md files when changing chart functionality"* and *"Document each configuration parameter in values.yaml"*.

The fix is to express the new surface as `@param` entries where it is parameterised (e.g. `jobs[].securityContext.*`, `cronjobs[].securityContext.*`) so the generator carries it, and hand-edit the CronJob table for the precedence rule, which has no parameter to hang off.

*Anchor note: `README.md` is not in the diff. These lines are the largest block of new prose the generator will not pick up, so the comment sits on its cause.*

charts/mycarrier-helm/tests/security_context_test.yaml
```diff
@@ -150,4 +150,46 @@ tests:
...
+          seccompProfile:
+            type: RuntimeDefault
+    asserts:
+      - isKind:
+          of: Deployment
+      - equal:
+          path: spec.template.spec.securityContext.fsGroup
+          value: 1000000
+      - equal:
+          path: spec.template.spec.securityContext.seccompProfile.type
```
smunukutla-mycarrier, line 176: **💡 suggestion · technical-reviewer · Logic**

### The applications `fsGroup` bound is untested, and no test covers the boundary value on any surface

`values.schema.json` gains `"minimum": 1` at three sites — applications (`:610`), jobs (`:1742`), cronjobs (`:2013`). `job_test.yaml:807` and `cronjob_test.yaml:885` each add a `failedTemplate` rejection test. There is no equivalent for `applications[].securityContext.fsGroup` here, in `rollout_test.yaml`, or in `statefulset_test.yaml`.

The constraint does fire — verified: `at '/applications/demo/securityContext/fsGroup': minimum: got 0, want 1` — it is simply unguarded against removal. The applications block matters most because it is one of the three entry points `mc-environment` remote-`$ref`s (`values.schema.json#/properties/applications`), so a regression there propagates to that chart's consumers immediately via the floating ref.

**Amended recommendation:** a fourth copy of the `fsGroup: 0` rejection test is the lower-value option. The genuinely missing coverage is the **boundary**. All four new fsGroup tests use `1000000`; nothing asserts that `fsGroup: 1` — the smallest value `"minimum": 1` admits, and the value that most directly exercises whether the guard is falsey-based or presence-based — actually renders. It does today:

```
cronjobs[0].securityContext.fsGroup: 1
-> securityContext: runAsUser: 1000 / runAsGroup: 3000 / runAsNonRoot: true / fsGroup: 1
```

A boundary test pins both the bound and the guard, and would fail if the `fsGroup` schema/guard changes discussed elsewhere in this review were implemented carelessly. Prefer one `fsGroup: 1` render assertion plus the applications-surface rejection test over two more copies of the existing one.

smunukutla-mycarrier (changes requested):

charts/mycarrier-helm/templates/_annotations.tpl
pkuzmenko-mycarrier: Fixed, and taken further than the comment asked. `userExtra` now filters every annotation surface: the three application pod templates, Job resource metadata, and the Job pod template.

On istio/gateway I went the other way, though — I reserved them too. The reasoning you gave was that the annotation is the only per-workload Istio escape hatch, but `_helpers.tpl:203` reads `applications[].istioDisabled`, and `helm.annotations.gateway` branches on `networking.istio.enabled` and `service.istioDisabled`. Those are first-class fields, so the annotation override was redundant — and leaving it open kept a real duplicate key alive: `sidecar.istio.io/inject` rendered twice on every Deployment where a user overrode it. Blast radius checked first — zero values files in the org override those three keys on an application.

Cronjobs and jobs are unaffected: their pod templates emit no istio annotations, so `sidecar.istio.io/inject: "false"` still works there and is still the documented way to stop a sidecar hanging a Job.

Verified: a values file engineered to collide on every surface now renders zero duplicate YAML keys across all 8 documents; the same input against main produces 3.

pkuzmenko-mycarrier: Fixed, and taken further than the comment asked. `userExtra` now filters every annotation surface: the three application pod templates, Job resource metadata, and the Job pod template.

On istio/gateway I went the other way, though — I reserved them too. The reasoning you gave was that the annotation is the only per-workload Istio escape hatch, but `_helpers.tpl:203` reads `applications[].istioDisabled`, and `helm.annotations.gateway` branches on `networking.istio.enabled` and `service.istioDisabled`. Those are first-class fields, so the annotation override was redundant — and leaving it open kept a real duplicate key alive: `sidecar.istio.io/inject` rendered twice on every Deployment where a user overrode it. Blast radius checked first — zero values files in the org override those three keys on an application.

Cronjobs and jobs are unaffected: their pod templates emit no istio annotations, so `sidecar.istio.io/inject: "false"` still works there and is still the documented way to stop a sidecar hanging a Job.

Verified: a values file engineered to collide on every surface now renders zero duplicate YAML keys across all 8 documents; the same input against main produces 3.

charts/mycarrier-helm/values.yaml
pkuzmenko-mycarrier: Correct, and this one was my error. The docs now list the exact per-surface reserved sets instead of claiming a `vault.security.banzaicloud.io/*` prefix, including the `manualOtelConfig` third gate you traced, and the fact that only `inject-<resolved-language>` is emitted rather than the whole `instrumentation.opentelemetry.io/*` namespace.

There is now an explicit warning that any unreserved key still applies to the pod, naming `vault-skip-verify`, `mutate: skip`, `vault-namespace` and `vault-serviceaccount`.

I also took your reasoning for withdrawing the blanket prefix-match rather than just your conclusion — reserving the whole namespace would trade silent-override for silent-drop on knobs the vendor documents, which is worse.

charts/mycarrier-helm/values.yaml
pkuzmenko-mycarrier: Half right, and the half that is wrong matters — so I checked the live clusters rather than the repo.

Each app cluster syncs its **own branch**. `perClusterRevision: true` has been enabled since 2026-07-24 (`53112cc`), and each cluster's live sync revision matches its branch HEAD exactly. `main` is deployed to nothing.

- **development cluster**: `restrict-seccomp` is `validationFailureAction: Enforce`, and the live pattern is `runtime/default | RuntimeDefault` (commit `d19f9de`). `RuntimeDefault` **passes**, and it is genuinely enforced at admission.
- **data / production-csp clusters**: Audit, pattern accepts only the literal `runtime/default`. `RuntimeDefault` records a PolicyReport entry and is not blocked.

So the branch you read is real, but no cluster runs it. Your core catch still stands for data/prod and the prose was wrong either way, so I have rewritten it per-cluster rather than adopting the suggestion, which would have been wrong for the development cluster.

Your PSS point I did confirm, and it corrected me: the Pod Security Standards page contains zero occurrences of `fsGroup`, `runAsGroup` or `supplementalGroups` — those were PodSecurityPolicy, retired in 1.25. The `>= 1` bound is now attributed to kyverno `require-non-root-groups` rule `check-fsGroup`, not to an upstream standard.

charts/mycarrier-helm/values.schema.json
pkuzmenko-mycarrier: Taken in full. `securityContext` now has `additionalProperties: false`, and `seccompProfile` requires `type` against `["RuntimeDefault","Localhost","Unconfined"]` with its own `additionalProperties: false`.

All three of your reproductions now fail at schema time:

```
readOnlyRootFileSystem: true      -> additional properties 'readOnlyRootFileSystem' not allowed
seccompProfile: {}                -> missing property 'type'
seccompProfile: {typo: ...}       -> missing property 'type' + additional properties 'typo' not allowed
seccompProfile: {type: Bogus}     -> value must be one of 'RuntimeDefault', 'Localhost', 'Unconfined'
```

I kept all three enum values rather than narrowing to PSS-restricted, on your API-grounds reasoning — the schema should not reject a manifest Kubernetes accepts.

On staging it rather than taking it now: I confirmed the floating `$ref` independently (see the thread on `values.yaml:112`) and went ahead anyway, because an org-wide search found zero consumers setting `securityContext` on applications, jobs or cronjobs — so the measured blast radius is zero even though the structural exposure is real.

charts/mycarrier-helm/templates/_annotations.tpl
pkuzmenko-mycarrier: Right on both counts, and the second was the more useful one.

Job resource metadata now derives its reserved set from a new `helm.annotations.jobMetadata`, mirroring `cronjobMetadata`. And your three-different-values observation is exactly why I did **not** try to unify them into a single define — `Prune=true`, `Force=true,Replace=true` and `Prune=true,Replace=true` are genuinely three different things, and the pod-template one has conditional hook/sync-wave keys on top.

For the Job pod template I reserved all four argocd keys unconditionally — `sync-options`, `hook-delete-policy`, `sync-wave`, `hook` — even though the last two only render when `.job.timing` is set. That keeps the reserved set independent of `.job.timing`, so enabling a hook later cannot silently change which user keys survive. There is a test pinning it.

Your Trivy reproduction on the Job surface matched mine.

charts/mycarrier-helm/templates/_spec_cronjob.tpl
pkuzmenko-mycarrier: Diagnosis accepted, remedy declined, and one claim does not hold up.

I reproduced the drift experiment — adding a fourth chart-managed helper without updating the reserved set does produce a live duplicate key, exactly as you describe. But `helm unittest -u` does **not** return the suite to green:

```
helm unittest -u  ->  Tests: 1 failed, 606 passed, 607 total
```

The exact-map `equal` assertion is a hardcoded literal in the test YAML, not a snapshot, so `-u` cannot touch it. A developer taking the natural shortcut still sees a red test.

On the remedy: building the chart-managed annotations as a dict and `toYaml`-ing it re-sorts the keys alphabetically and rewrites the `vault-agent-env-variables` folded scalar (`>-`) into a quoted string. That is semantically identical but not byte-identical, and byte-identical default rendering is a guarantee this PR makes and verifies 6/6 against main. That is why the implementation is filter-then-append rather than a dict round-trip — the chart's own annotation output is never reparsed or re-emitted.

Your point about the hand-maintained list is still valid; I have not solved it, I have traded it for a property I care about more.

charts/mycarrier-helm/templates/_annotations.tpl
pkuzmenko-mycarrier: Fair, and I am accepting the capability point without adding the field.

Chart-wins does close off the (accidental) route by which someone could have appended `ServerSideApply=true` or `PruneLast=true` at CronJob metadata. I have documented that the field is now closed rather than inventing `cronjobs[].syncOptions` with no consumer asking for it — that is speculative API surface, and this chart already has more knobs than documented behaviour.

Your `Prune=true` observation checks out: the ArgoCD sync-options page documents `Prune=false` and `Prune=confirm`, and `Prune=true` is not among them. That is worth its own look, independent of this PR — it predates it on every surface.

Thanks for withdrawing the CronJob immutability reasoning explicitly; it saved me chasing it.

charts/mycarrier-helm/values.schema.json
pkuzmenko-mycarrier: Fixed properly rather than papered over.

Both guards are now presence-based: `{{- if hasKey $sc "fsGroup" }}` and `{{- if hasKey $sc "seccompProfile" }}`. You were right that it was a class and not an instance — `seccompProfile: {}` had the identical defect, and it is now a schema rejection as well as no longer silently vanishing.

`"minimum": 1` stays, so `fsGroup: 0` never reaches the template. The `int64` cast stays too — it is guarded by a test that provably fails without it.

On attribution: I confirmed your PSS point and it corrected me. The bound is org policy from `require-non-root-groups`/`check-fsGroup`, and the schema `description` now says so, so the next reader knows where it came from.

charts/mycarrier-helm/values.yaml
pkuzmenko-mycarrier: The `$ref` half is the real finding here, and I verified it end to end rather than taking it on the citation.

All 23 refs in `charts/mc-environment/values.schema.json` float on `refs/heads/main`, and Helm genuinely fetches and enforces them — rendering mc-environment with an invalid value fails against the *remote* enum, not a local one. So schema edits in this repo are effectively released to mc-environment on merge, with `mycarrierChartVersion` doing nothing to gate them. You are right that this inverts the usual remedy.

I went ahead with the tightening anyway, on measured blast radius: an org-wide search found zero consumers setting `securityContext` on applications, jobs or cronjobs, so nothing breaks on merge. That was an explicit call, not an oversight.

Pinning the `$ref` is a genuine defect but it changes how every mc-environment consumer resolves schemas, in a different chart, so it is not riding along in this PR.

On the changelog: there is no CHANGELOG in this repo to add to. The three behaviour changes are described in the PR body and in `values.yaml`.

charts/mycarrier-helm/templates/cronjob.yaml
pkuzmenko-mycarrier: Confirmed, and removed.

`fromYaml` returning `{"Error": ...}` on failure means `| default dict` cannot fire on the error branch, and on the empty-input branch it substitutes an equally empty dict — dead both ways, exactly as you said. It is gone from all seven reserved-set sites.

Rather than the dict refactor (see my reply on `_spec_cronjob.tpl:51-54` for why that one breaks byte-identity), I took your alternative: a single fail-fast guard, placed inside `helm.annotations.userExtra` so all seven call sites inherit it rather than repeating a check seven times. It fires only on chart-authored parse failures, never on user collisions, with a comment stating that distinction so nobody later generalises it.

Proven to fire:

```
Error: execution error at (mycarrier-helm/templates/jobs.yaml:25:16): helm.annotations.userExtra: chart-managed annotations failed to parse: error converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context
```

On your `Error`-keyed-user-annotation caveat: the reserved map is built exclusively from chart-defined helpers, and user annotations are passed separately as `user` and only ever read via `hasKey $reserved $k`. A user key cannot enter `$reserved`, so it cannot trip the guard.

charts/mycarrier-helm/templates/_podsecurity.tpl
pkuzmenko-mycarrier: Taken as-is — it is now an exclusive `if / else if / else if` chain matching `_labels.tpl`.

Your corrected call-site count of 15 was right, and so was the `_labels.tpl` correction: it is the counter-example, not the precedent, which is what made the two-word fix the obviously right size of change here.

charts/mycarrier-helm/values.yaml
pkuzmenko-mycarrier: Closed rather than documented.

`jobs[].annotations` now render onto the Job pod template as well as the Job resource, filtered like every other surface. Your asymmetry argument is what decided it: a hung PreSync hook blocks the entire application sync, which is strictly worse than a hung CronJob-spawned pod blocking one scheduled run. Documenting a gap that severe would have been the wrong call when the fix is the same helper already in the diff.

I also confirmed your substantiation — `applications[].istioDisabled` is read via `.application` and genuinely cannot reach a `jobs[]` item, since `jobs[]` is a top-level array with no application context. There was no workaround.

`_spec_triggertestengine.tpl` I have left alone: it has no annotation input at all, so closing that gap means adding new API surface with no values key behind it. Worth doing, but not silently in this PR.

charts/mycarrier-helm/values.yaml
pkuzmenko-mycarrier: Right about the effect, and I tried your fix before rejecting it.

I cloned `bitnami/readme-generator-for-helm@2.7.2` and ran it against test fixtures. Adding `## @param jobs[].securityContext.*` (or the cronjobs/applications equivalents) throws:

```
Error: ERROR: Wrong metadata!
```

and exits non-zero. The tool requires every `@param` path to resolve to a real key in the parsed values AST, and `jobs`/`cronjobs` default to `[]` while `applications` defaults to `{}` — there is no array-item node to attach a nested path to. All 41 existing `@param` lines in this chart document concrete, non-empty keys; none document array-item shapes. That is a tool limitation, not an oversight, and following the suggestion literally would have broken `readme-gen.yaml` on the next PR touching `values.yaml`.

So I did the hand-written half instead, which was the substance of your point: the CronJob Configuration Options table now has `annotations`, `labels` and `securityContext` rows, and there is a new "Security Context and Reserved Annotations" section covering the keys, the per-cluster kyverno seccomp situation, and the precedence rule. `grep -c securityContext README.md` was 0; it is not now.

Thanks for withdrawing the helper-function limb — chasing 141 defines would have been wasted effort.

charts/mycarrier-helm/values.schema.json
pkuzmenko-mycarrier: Taken in full — the three copies are now one `#/definitions/workloadSecurityContext`, referenced from applications, jobs and cronjobs.

Draft-07 dialect, so `definitions` rather than `$defs` (the file declares `"$schema": "https://json-schema.org/draft-07/schema#"`). Both drift items you named are fixed by construction: the `readOnlyRootFilesystem` description divergence is gone, and `addCapabilities` is declared for the first time — with a note that `_podsecurity.tpl` only honours it on the application path, which is worth knowing before someone sets it on a cronjob.

`additionalProperties` is now one decision across all three, as you suggested, and it is `false` — see the thread on `:1744` for the blast-radius reasoning.

Your remote-`$ref` test matched mine: mc-environment refs sub-paths of this file and still lints and templates cleanly after the extraction, so the local pointer resolves against the remote base URI as the spec requires.

charts/mycarrier-helm/tests/security_context_test.yaml
pkuzmenko-mycarrier: Both added, and I took your amendment over the literal ask.

There is now an applications-surface `fsGroup: 0` rejection test *and* `fsGroup: 1` boundary cases. You were right that a fourth copy of the rejection test was the lower-value option — the boundary is what actually distinguishes a falsey guard from a presence guard.

That turned out to matter more than either of us knew at the time: I changed the guard from `{{- with $sc.fsGroup }}` to `{{- if hasKey $sc "fsGroup" }}` in this same round, per the thread on `values.schema.json:610`. The `fsGroup: 1` test is what pins that behaviour going forward.

hpyvovar-mycarrier (approved):

smunukutla-mycarrier (approved):

PR #112, opened 2026-08-13
feat(mycarrier-helm): default testtrigger hardenedSecurityContext to true

pkuzmenko-mycarrier:
## Summary
- Flips the testtrigger `hardenedSecurityContext` default from `false` to `true` (DEVOPS-176 Lever 1) — test jobs are now hardened (runAsNonRoot + readOnlyRootFilesystem via TestEngine) unless a service explicitly opts out.
- Explicit `hardenedSecurityContext: false` in a testdefinition is still honored — the template already distinguishes absent-vs-false via `hasKey`, only the initializer changed (both the v1 and legacy payload paths).

## Changes
- **Template**: `_spec_triggertestengine.tpl` lines 92/127 — initializer `false` → `true` at both payload sites; nothing else touched.
- **Tests**: +5 helm-unittest cases in `triggertests_test.yaml` asserting absent→`true`, explicit `true`→`true`, explicit `false`→`false` on both paths.
- **Chart.yaml**: 3.5.8 → 3.6.0 (behavior-changing default = minor; CI version-bumper skips auto-patch when manually bumped).

## Render proof (3-way, both payload paths)
| testdefinition key | Before | After |
|---|---|---|
| absent | `"HardenedSecurityContext": false` | `"HardenedSecurityContext": true` |
| explicit `true` | `true` | `true` |
| explicit `false` | `false` | `false` (opt-out honored) |

## How to verify
1. `helm unittest charts/mycarrier-helm -f 'tests/triggertests_test.yaml'` — 43/43 pass (full suite 628/628, was 623/623 baseline).
2. `helm lint charts/mycarrier-helm` — 0 failures.
3. `helm template` a values file with a testdefinitions entry lacking the key — rendered curl payload shows `"HardenedSecurityContext": true`.

## Rollout notes (pre-merge gate)
- 5 producers currently lack the key and will be force-hardened on this release: `MC.MyCarrier.Frontend-new` (all envs), `MC.Analytics`/`MC.AngularExample`/`MC.CatalystPOC` (prod values only — dev/preprod already explicit `true`), `exampletestapp` (dev). None are write-path-validated; suites writing outside TestEngine's writable mounts would hit EROFS. Either validate those suites or land explicit `false` opt-outs in those values files before bumping consumer pins.
- `values.yaml` line 585 comment still documents `default: false` — follow-up doc fix.
- Rollback lever is per-service explicit `false` (proven honored) — no chart revert needed.

## Checklist
- [x] Chart tests pass (628/628, incl. 5 new)
- [x] Lint passes
- [x] Before/after render proof captured
- [x] Chart version bumped (minor)
- [x] No secrets committed

### DIFF
--- a/charts/mycarrier-helm/Chart.yaml
+++ b/charts/mycarrier-helm/Chart.yaml
@@ -3,5 +3,5 @@ name: mycarrier-helm
 description: A Helm chart for MyCarrier GitOps
 type: application
 icon: https://go.mycarrier.io/hs-fs/hubfs/Logos/PNG-MC-COLOR-Logo.png?width=366&height=85&name=PNG-MC-COLOR-Logo.png
-version: 3.5.8
+version: 3.6.0
 appVersion: 3.0.29
--- a/charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
+++ b/charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
@@ -89,7 +89,7 @@ template:
           {{- if hasKey . "spreadAcrossNodes" }}
             {{- $spreadAcrossNodes = printf "%v" .spreadAcrossNodes }}
           {{- end }}
-          {{- $hardenedSecurityContext := false }}
+          {{- $hardenedSecurityContext := true }}
           {{- if hasKey . "hardenedSecurityContext" }}
             {{- $hardenedSecurityContext = .hardenedSecurityContext }}
           {{- end }}
@@ -124,7 +124,7 @@ template:
           {{- if hasKey . "spreadAcrossNodes" }}
             {{- $spreadAcrossNodes = printf "%v" .spreadAcrossNodes }}
           {{- end }}
-          {{- $hardenedSecurityContext := false }}
+          {{- $hardenedSecurityContext := true }}
           {{- if hasKey . "hardenedSecurityContext" }}
             {{- $hardenedSecurityContext = .hardenedSecurityContext }}
           {{- end }}
--- a/charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
+++ b/charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
@@ -90,8 +90,8 @@ template:
             {{- $spreadAcrossNodes = printf "%v" .spreadAcrossNodes }}
           {{- end }}
           {{- $hardenedSecurityContext := true }}
-          {{- if hasKey . "hardenedSecurityContext" }}
-            {{- $hardenedSecurityContext = .hardenedSecurityContext }}
+          {{- if and (hasKey . "hardenedSecurityContext") (eq (lower (printf "%v" .hardenedSecurityContext)) "false") }}
+            {{- $hardenedSecurityContext = false }}
           {{- end }}
           {{- $defaultPodResources := dict "limits" (dict "cpu" "2000m" "memory" "4Gi") "requests" (dict "cpu" "250m" "memory" "0.5Gi") }}
           {{- $podResources := .podResources | default $defaultPodResources }}
@@ -125,8 +125,8 @@ template:
             {{- $spreadAcrossNodes = printf "%v" .spreadAcrossNodes }}
           {{- end }}
           {{- $hardenedSecurityContext := true }}
-          {{- if hasKey . "hardenedSecurityContext" }}
-            {{- $hardenedSecurityContext = .hardenedSecurityContext }}
+          {{- if and (hasKey . "hardenedSecurityContext") (eq (lower (printf "%v" .hardenedSecurityContext)) "false") }}
+            {{- $hardenedSecurityContext = false }}
           {{- end }}
           {{- $defaultPodResources := dict "limits" (dict "cpu" "2000m" "memory" "4Gi") "requests" (dict "cpu" "250m" "memory" "0.5Gi") }}
           {{- $podResources := .podResources | default $defaultPodResources }}
--- a/charts/mycarrier-helm/tests/triggertests_test.yaml
+++ b/charts/mycarrier-helm/tests/triggertests_test.yaml
@@ -975,38 +975,6 @@ tests:
           path: spec.template.spec.containers[0].command[2]
           pattern: '"HardenedSecurityContext":\s*false'
 
-  - it: should set HardenedSecurityContext to true when explicitly specified (legacy mode)
-    set:
-      fullnameOverride: app
-      namespace: dev-app
-      metaEnvironment: dev
-      applications:
-        app1:
-          version: 1.0.0
-          enabled: true
-          image:
-            registry: "myregistry.example.com"
-            repository: "mycarrier/app1"
-            tag: "1.0.0"
-          testtrigger:
-            ttlSecondsAfterFinished: "300"
-            testdefinitions:
-              - name: "apitests"
-                containerImage: "alpine/curl"
-                containerTag: "latest"
-                secretId: "vault:Secret/data/secret#SecretId"
-                filters:
-                  - "TestCategory=whatever"
-                hardenedSecurityContext: true
-    asserts:
-      - isKind:
-          of: Job
-      - hasDocuments:
-          count: 1
-      - matchRegex:
-          path: spec.template.spec.containers[0].command[2]
-          pattern: '"HardenedSecurityContext":\s*true'
-
   - it: should include HardenedSecurityContext as true by default when enableV1 is true
     set:
       fullnameOverride: app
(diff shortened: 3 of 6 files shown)


charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
```diff
@@ -89,7 +89,7 @@ template:
           {{- if hasKey . "spreadAcrossNodes" }}
             {{- $spreadAcrossNodes = printf "%v" .spreadAcrossNodes }}
           {{- end }}
-          {{- $hardenedSecurityContext := false }}
+          {{- $hardenedSecurityContext := true }}
           {{- if hasKey . "hardenedSecurityContext" }}
             {{- $hardenedSecurityContext = .hardenedSecurityContext }}
           {{- end }}
```
hpyvovar-mycarrier, line 95: 🟠 **important — Opt-out value is not normalized; nil/string/typo break or diverge the payload** (Contract/Logic)

`.hardenedSecurityContext` is assigned verbatim at line 94 (and identically at 129) with no coercion — it's the one sibling flag not normalized via `printf "%v"` (compare `spreadAcrossNodes` :90, `useDefaultNodeAffinity` :117, `legacyMode` :121). This PR makes the explicit key the **sole** opt-out, so three input shapes now break or silently diverge (all verified with `helm template` at this head SHA):

- `hardenedSecurityContext:` (null) → legacy renders `"HardenedSecurityContext": ,` = **invalid JSON**; the POST is rejected and — because the curl calls carry no `--fail` — the Job still reports Success (false-green, zero tests run). v1 renders `null`.
- `hardenedSecurityContext: "false"` (string — the schema tells producers strings are acceptable for the sibling flags, and this suite already uses the quoted form at :709/:804/:1128) → **v1 emits a JSON string `"false"`, legacy emits a bare `false`**. The two payload paths disagree on the wire type for identical input; a strict bool deserializer on the v1 side would reject or ignore the opt-out — silently, in the hardened direction.
- `hardenedSecurityContext: falsee` (typo) → legacy invalid JSON, v1 `"falsee"`.

Fail-secure fix below — accepts bool **and** case-insensitive string, only a literal `false` opts out, nil/garbage stay `true`, and it emits a real Go bool so both payload paths agree. **Apply the same change at both sites (:92-95 and :127-130).**

```suggestion
          {{- $hardenedSecurityContext := true }}
          {{- if and (hasKey . "hardenedSecurityContext") (eq (lower (printf "%v" .hardenedSecurityContext)) "false") }}
            {{- $hardenedSecurityContext = false }}
          {{- end }}
```

(Note: both agents' first-pass patches were rejected in review — `kindIs "bool"` silently drops the in-use string form, and `printf "%v" == "true"` fails *open*, mapping `falsee`/`no`/`0` to unhardened.)

charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
```diff
@@ -89,7 +89,7 @@ template:
           {{- if hasKey . "spreadAcrossNodes" }}
             {{- $spreadAcrossNodes = printf "%v" .spreadAcrossNodes }}
           {{- end }}
-          {{- $hardenedSecurityContext := false }}
+          {{- $hardenedSecurityContext := true }}
           {{- if hasKey . "hardenedSecurityContext" }}
             {{- $hardenedSecurityContext = .hardenedSecurityContext }}
```
hpyvovar-mycarrier: 🟠 **important — `values.schema.json` never declares `hardenedSecurityContext`** (Contract)

(Fix belongs in `charts/mycarrier-helm/values.schema.json`; anchored here to the line that consumes the value.)

The `testdefinitions.items` schema declares 13 properties and `hardenedSecurityContext` is not among them, and there is no `additionalProperties: false`, so `helm install` accepts **any** type for it — `null`, `"false"`, `falsee` — which is exactly what lets the broken input shapes in the previous comment reach a rendered manifest instead of failing at validation time. Every sibling boolean is declared `{"type": ["string","boolean"], "default": ...}`.

Add to `testdefinitions.items.properties`:
```json
"hardenedSecurityContext": { "type": ["string", "boolean"], "description": "Run test pods with a hardened security context. Accepts boolean or string.", "default": true }
```
This rejects `null`. It won't catch the `falsee` typo (a valid string), so it **complements** the normalization fix above rather than replacing it.

charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
```diff
@@ -89,7 +89,7 @@ template:
           {{- if hasKey . "spreadAcrossNodes" }}
             {{- $spreadAcrossNodes = printf "%v" .spreadAcrossNodes }}
           {{- end }}
-          {{- $hardenedSecurityContext := false }}
+          {{- $hardenedSecurityContext := true }}
```
hpyvovar-mycarrier, line 92: 🟡 **suggestion — Chart docs still state the old default; the guide has no row at all** (Convention)

(Fixes belong in `values.yaml` and `TestTriggersGuide.md`; anchored to the changed initializer.)

- `charts/mycarrier-helm/values.yaml:585` still reads `hardenedSecurityContext: false          # Optional (default: false)` — now the opposite of the shipped behavior. (You disclosed this in the PR body as a follow-up; flagging so it lands in the same release.)
- `charts/mycarrier-helm/TestTriggersGuide.md` (option table, lines 98-113) has **no `hardenedSecurityContext` row at all**, while every sibling flag is documented with a Default column. A behavior now on-by-default for every test pod is undocumented in the guide producers are pointed at.

Both are one-line edits.

charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
```diff
@@ -124,7 +124,7 @@ template:
...
           {{- if hasKey . "hardenedSecurityContext" }}
             {{- $hardenedSecurityContext = .hardenedSecurityContext }}
           {{- end }}
```
hpyvovar-mycarrier, line 130: 🟠 **important (pre-existing — fast-follow, not a merge blocker) — Shell + JSON injection in the curl payload** (Security-Injection)

Not introduced by this PR — anchored to the nearest changed line. The real sinks are the `curl … -d '…'` invocations at :105 (v1) and :133-160 (legacy), and the raw-interpolated fields at :142/:155/:156.

The payload is embedded in a **single-quoted argument to `/bin/sh -c`**. A single quote in any interpolated value breaks out and runs arbitrary commands in the trigger pod, which carries `TESTENGINE_APIKEY` in-env and mounts the `default` ServiceAccount token (`readOnlyRootFilesystem: false`, so `wget | sh` works). Additionally the **legacy path applies no JSON escaping** to `.name` / `.containerImage` / `.secretId` / `$gitBranch` / `.additionalEnvVars` — verified end-to-end that a `"` in `additionalEnvVars` injects a second `"HardenedSecurityContext": false` into valid JSON, and last-key-wins (System.Text.Json and Newtonsoft) → **silently defeats the exact hardening this PR adds**. `$gitBranch` derives from `global.gitbranch`, and git permits quotes in ref names.

Hardened fix (fast-follow ticket): deliver the payload over stdin with a quoted heredoc (`-d @-  … <<'PAYLOAD'`) instead of a single-quoted arg, and pass every legacy scalar through `toJson` (as v1 already does).

hpyvovar-mycarrier, line 130: 🟡 **suggestion — v1 and legacy carry a byte-identical 25-line defaulting prologue edited in lockstep** (DRY)

Lines 73-97 (v1) and 108-132 (legacy) are byte-for-byte identical (`diff` of the two ranges → zero differences); only the serialization below them legitimately differs (`toPrettyJson` at :105 vs. the hand-written JSON at :136-160). This PR is the evidence of the cost — one semantic decision ("default is now true") required two identical edits in two hunks, and the normalization fix above has to land twice too. Nothing catches a future editor changing one site and not the other; the new tests assert v1 and legacy independently and never assert they **agree**.

Cheapest mitigation (no refactor): add one helm-unittest case asserting v1 and legacy resolve identically for the same input. (A `fromJson`-backed `define` would collapse both sites but risks coercing the structured `$tolerations`/`$nodeAffinity`/`$podResources` values through a JSON round-trip — scope it to the scalar flags if you take that route.)

charts/mycarrier-helm/tests/triggertests_test.yaml
```diff
@@ -912,6 +912,166 @@ tests:
...
+                  - "TestCategory=whatever"
+                hardenedSecurityContext: false
+    asserts:
+      - isKind:
+          of: Job
+      - hasDocuments:
+          count: 1
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: '"HardenedSecurityContext":\s*false'
```
hpyvovar-mycarrier, line 976: 🟡 **suggestion — The explicit-`true` case has no discriminating power** (Logic)

Now that `true` is the default, asserting `"HardenedSecurityContext":\s*true` on explicit-`true` input passes whether or not the `hasKey` override branch runs — it can't distinguish "my explicit value was honored" from "I got the default." Confirmed by mutation: deleting **both** `hasKey` override branches fails only the two explicit-`false` cases; this case and both default cases still pass.

Of the 5 new cases, 2 carry the regression signal (the default cases — they'd fail on `main`) and 3 are confirmatory. This one buys nothing. Better: replace it with the null / string-`"false"` / `falsee` cases from the normalization finding — asserted against the **fixed** behavior, not today's buggy renders — so the suite actually guards the opt-out's type handling.

charts/mycarrier-helm/Chart.yaml
```diff
@@ -3,5 +3,5 @@ name: mycarrier-helm
 description: A Helm chart for MyCarrier GitOps
 type: application
 icon: https://go.mycarrier.io/hs-fs/hubfs/Logos/PNG-MC-COLOR-Logo.png?width=366&height=85&name=PNG-MC-COLOR-Logo.png
-version: 3.5.8
+version: 3.6.0
```
hpyvovar-mycarrier, line 6: 🟠 **important — Rollout gate: 5 producers force-hardened without write-path validation** (the review's top open question)

Anchored to the version bump because that's what makes this the consumable release. Per the PR body, 5 producers lack the key and will be force-hardened on 3.6.0 — `MC.MyCarrier.Frontend-new`, `MC.Analytics`, `MC.AngularExample`, `MC.CatalystPOC`, `exampletestapp` — none write-path-validated; a suite writing outside TestEngine's writable mounts hits `EROFS` and starts failing.

This is a rollout precondition, not a diff defect, but it's the most consequential item here. Confirm the pre-merge gate you set yourself — *"validate those suites or land explicit `false` opt-outs before bumping consumer pins"* — is satisfied or tracked before this tag is consumed. (Chart version bump 3.5.8 → 3.6.0 itself is correct and matches repo precedent `f0b96d53`.)

hpyvovar-mycarrier (changes requested):

charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
pkuzmenko-mycarrier: Fixed in ba77078 — applied your suggested coercion verbatim at both sites (lines 92-95 and 127-130). Only a literal `false` or case-insensitive string `"false"` opts out now; nil, typos, and any other type stay hardened, and the variable is always a real Go bool so the v1 and legacy payloads agree on the wire type. The null render-break is additionally unreachable now that the schema rejects null (next thread).

charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
pkuzmenko-mycarrier: Fixed in ba77078 — `hardenedSecurityContext` is now declared in `testdefinitions.items.properties` as `{"type": ["string","boolean"], "default": true}` with a description, matching the sibling flags. Verified empirically: `helm template` with a null value now fails schema validation ("got null, want boolean or string") on both payload paths, and two `failedTemplate` unit tests pin that behavior. As you noted it cannot catch the `falsee` typo, so it complements the template-level coercion rather than replacing it.

charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
pkuzmenko-mycarrier: Fixed in ba77078 — `values.yaml:585` example now reads `hardenedSecurityContext: true # Optional (default: true; set false to opt out of hardened test pods)`, and `TestTriggersGuide.md` gained a `hardenedSecurityContext` row in the option table (Default `true`, opt-out via `false`), so both land in the same release as the behavior change.

charts/mycarrier-helm/templates/_spec_triggertestengine.tpl
pkuzmenko-mycarrier: Acknowledged — agreed this is pre-existing and out of scope for this PR, and it does deserve a fast follow given the interaction you found where an injected duplicate key can silently defeat the hardening this PR turns on. Filed as tracker issue mycarrier-14pk (P1) with your fix direction verbatim: deliver the payload over stdin with a quoted heredoc (`-d @- <<'PAYLOAD'`) and pass every legacy scalar through `toJson` as the v1 path already does.

pkuzmenko-mycarrier: Took the cheapest mitigation in ba77078: the new type-handling tests assert the exact rendered `"HardenedSecurityContext"` value on BOTH the v1 and legacy paths for every input shape (absent, explicit false, string "false", "FALSE", typo `falsee`, null) — so a future editor changing one prologue site and not the other now fails the suite on the first shape where the two paths disagree. The `fromJson`-backed `define` collapse is deferred as you suggested, given the coercion risk to the structured `$tolerations`/`$nodeAffinity`/`$podResources` values.

charts/mycarrier-helm/tests/triggertests_test.yaml
pkuzmenko-mycarrier: Fixed in ba77078 — the explicit-`true` case is removed and replaced with the cases that actually carry signal against the fixed coercion: quoted `"false"` → false and `"FALSE"` → false (opt-out honored for the in-use string form, case-insensitive), typo `falsee` → stays true (fail-secure), and null → schema rejection pinned via `failedTemplate` tests. All asserted on both payload paths. Suite went from 628 to 635 tests, all passing.

charts/mycarrier-helm/Chart.yaml
pkuzmenko-mycarrier: Acknowledged — this is the release gate and it is deliberately NOT resolved inside this PR. Current state: an independent per-testdefinition sweep (195 entries across 165 values files) confirmed the 5 producers you listed are the complete unvalidated set, and nothing anywhere sets an explicit `false` today. The gate before any consumer pin bump: either write-path-validate those 5 suites or land explicit `false` opt-outs in their values files (the opt-out path is now proven honored, including the string form). Merging this chart change alone does not harden anyone until pins are bumped, and that step stays behind this precondition.

hpyvovar-mycarrier (approved):

PR #115, opened 2026-08-17
docs(mycarrier-helm): document disableSecurity break-glass flag + schema entry

pkuzmenko-mycarrier:
## Context
DEVOPS-176 follow-up (bd mycarrier-c463.31). After the `enableVaultCA` guard removal in `_podsecurity.tpl` (bd homc), `disableSecurity` is the SOLE flag that ships pods without `runAsNonRoot`/securityContext hardening. It was undocumented and absent from `values.schema.json`.

## Change
Doc + schema ONLY, zero behavior change.

- `values.schema.json`: property added following the `enableVaultCA` sibling pattern (`oneOf` boolean/null, default `false`, break-glass description) — no `additionalProperties`/`required` changes, so it cannot reject existing valid consumer values.
- `values.yaml`: key `disableSecurity: false` with break-glass comment.
- `README.md`: Platform Settings table row.

## Verification
- `helm lint`: PASS
- Default-values render: byte-identical before vs after
- `--set disableSecurity=true`: renders with hardening absent (expected break-glass behavior, unchanged)
- `--set-string disableSecurity=yes`: now FAILS schema validation ("got string, want boolean")
- `helm unittest`: 638/638 pass

## Notes
- The top-level schema has no `additionalProperties: false`, so a typo'd key name still passes silently — pre-existing, intentionally NOT tightened in this change (tightening it could reject existing consumer values).
- Chart.yaml version bump intentionally omitted — CI (`chart-version-bumper` via `.github/workflows/main.yml`) bumps + publishes on PR merge. Chart PUBLISH/propagation to consumers is a SEPARATE step from this merge.

### DIFF
--- a/charts/mycarrier-helm/values.schema.json
+++ b/charts/mycarrier-helm/values.schema.json
@@ -94,7 +94,7 @@
     },
     "disableSecurity": {
       "oneOf": [
-        { "type": "boolean", "description": "BREAK-GLASS ONLY (discouraged): when true, pods render WITHOUT runAsNonRoot/securityContext hardening. Sole remaining un-harden path after the enableVaultCA guard removal. Leave false/unset in normal operation.", "default": false },
+        { "type": "boolean", "description": "BREAK-GLASS ONLY (discouraged): when true, EVERY pod in the release renders WITHOUT runAsNonRoot/securityContext hardening — release-wide, cannot be scoped per-application. The only flag that removes the securityContext hardening block (applications.<name>.enableDebugMode's privileged sidecar is a separate un-hardened-container route). Kyverno require-run-as-non-root backstop is Enforce only on development; production-csp/data run Audit. Not settable via mc-environment environments[]. Leave false/unset; any use must be temporary.", "default": false },
         { "type": "null" }
       ]
     },
--- a/charts/mycarrier-helm/values.schema.json
+++ b/charts/mycarrier-helm/values.schema.json
@@ -92,6 +92,12 @@
         { "type": "null" }
       ]
     },
+    "disableSecurity": {
+      "oneOf": [
+        { "type": "boolean", "description": "BREAK-GLASS ONLY (discouraged): when true, pods render WITHOUT runAsNonRoot/securityContext hardening. Sole remaining un-harden path after the enableVaultCA guard removal. Leave false/unset in normal operation.", "default": false },
+        { "type": "null" }
+      ]
+    },
     "manualOtelConfig": {
       "oneOf": [
         { "type": "boolean", "description": "Take over OpenTelemetry configuration from the chart. When true the chart injects no OTEL_* env vars, no auto-instrumentation annotations and no otel-log volume, and overrides disableOtelAutoinstrumentation", "default": false },
--- a/charts/mycarrier-helm/values.yaml
+++ b/charts/mycarrier-helm/values.yaml
@@ -68,11 +68,17 @@ environment:
 ## Platform-wide settings
 
 ## @param enableVaultCA Enable Vault CA certificate download during pod startup (no longer affects securityContext hardening; that postStart hook is currently dead code)
+## @param disableSecurity BREAK-GLASS ONLY (discouraged): when true, pods render without runAsNonRoot/securityContext hardening
 ## @param manualOtelConfig Take over OpenTelemetry configuration from the chart (default false = chart-managed)
 ## @param disableOtelAutoinstrumentation Disable OpenTelemetry Operator auto-instrumentation only (set to false to enable)
 ## @param tolerations [array] Default tolerations for all applications
 ## @param deployment Deployment folder name (should match the subfolder containing the specific deployment)
 enableVaultCA: false
+
+## BREAK-GLASS ONLY — leave false. When true, pods ship WITHOUT runAsNonRoot/securityContext
+## hardening (both pod- and container-level). This is the sole remaining un-harden path since
+## the enableVaultCA guard removal; setting it is discouraged and should be temporary.
+disableSecurity: false
 
 ## The two OTel keys below have DELIBERATELY OPPOSITE defaults. They are not interchangeable and do
 ## not track each other:
--- a/charts/mycarrier-helm/values.yaml
+++ b/charts/mycarrier-helm/values.yaml
@@ -68,16 +68,23 @@ environment:
 ## Platform-wide settings
 
 ## @param enableVaultCA Enable Vault CA certificate download during pod startup (no longer affects securityContext hardening; that postStart hook is currently dead code)
-## @param disableSecurity BREAK-GLASS ONLY (discouraged): when true, pods render without runAsNonRoot/securityContext hardening
+## @param disableSecurity BREAK-GLASS ONLY (discouraged): when true, every pod in the release renders without runAsNonRoot/securityContext hardening
 ## @param manualOtelConfig Take over OpenTelemetry configuration from the chart (default false = chart-managed)
 ## @param disableOtelAutoinstrumentation Disable OpenTelemetry Operator auto-instrumentation only (set to false to enable)
 ## @param tolerations [array] Default tolerations for all applications
 ## @param deployment Deployment folder name (should match the subfolder containing the specific deployment)
 enableVaultCA: false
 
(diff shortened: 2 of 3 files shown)


charts/mycarrier-helm/values.yaml
```diff
@@ -68,12 +68,18 @@ environment:
...
 ## @param manualOtelConfig Take over OpenTelemetry configuration from the chart (default false = chart-managed)
 ## @param disableOtelAutoinstrumentation Disable OpenTelemetry Operator auto-instrumentation only (set to false to enable)
 ## @param tolerations [array] Default tolerations for all applications
 ## @param deployment Deployment folder name (should match the subfolder containing the specific deployment)
 enableVaultCA: false
 
+## BREAK-GLASS ONLY — leave false. When true, pods ship WITHOUT runAsNonRoot/securityContext
+## hardening (both pod- and container-level). This is the sole remaining un-harden path since
+## the enableVaultCA guard removal; setting it is discouraged and should be temporary.
+disableSecurity: false
```
hpyvovar-mycarrier: **Break-glass doc omits that the Kyverno backstop is dev-only — the flag is unbackstopped exactly where break-glass would be used** *(security · important)*

The added documentation is accurate as far as it goes, but omits three facts that determine this flag's real risk profile:

1. **The admission backstop is dev-only.** Verified in `kyverno-policies`: `require-run-as-nonroot` (and `deny-privilege-escalation`) are `Enforce` on `development` but `Audit` on `production-csp` and `data` (`main` tracks only the management cluster and runs no mycarrier-helm workloads). So on dev the flag is effectively inert — admission denies the unhardened pods — while on prod/data it takes full effect with report-only admission and no backstop. The chart's own comment (`_podsecurity.tpl:27`) hints at this; the user-facing docs should state where the safety net actually is.
2. **Release-wide blast radius.** `disableSecurity` is consumed only at `_podsecurity.tpl:30,54`, and those helpers are included by every workload spec (deployment, rollout, statefulset, cronjob, job, triggertestengine). `true` unhardens **all** pods in the release; it cannot be scoped per-application.
3. **"Sole remaining un-harden path" overclaims.** It is the only flag that removes the securityContext hardening block, but not the only route to an un-hardened container: `applications.<name>.enableDebugMode` injects a `privileged: true` sidecar on the Deployment path (`_debug.tpl`, wired at `_spec_deployment.tpl:187`). The same wording appears at `values.schema.json:97` and `README.md:1402` — worth narrowing all three to "the only flag that removes the securityContext hardening block".

Follow-ups worth separate issues (not this PR): unbounded `securityContext.addCapabilities` (`_podsecurity.tpl:66-71` — no allow-list, and no Kyverno capabilities policy on any branch; note it sits *inside* the `if not .Values.disableSecurity` block, so it is orthogonal to this flag), the chart-default `readOnlyRootFilesystem: false`, and a Kyverno Enforce graduation plan for prod/data so this flag regains a backstop.

For the record, this PR is a net security improvement: at the parent commit, `disableSecurity: "false"` (YAML string) silently disabled all hardening via Go-template truthiness with no error; the schema entry now rejects strings and ints.

```suggestion
## BREAK-GLASS ONLY — leave false. When true, EVERY pod in the release (all workloads) ships
## WITHOUT runAsNonRoot/securityContext hardening; this cannot be scoped per-application.
## Admission backstop: Kyverno require-run-as-non-root is Enforce ONLY on development —
## production-csp/data run Audit, so setting this there removes the chart's non-root guarantee
## with report-only admission and no backstop (containers run as whatever the image USER
## specifies). This is the only flag that removes the hardening block (though not the only
## route to an un-hardened container — see applications.<name>.enableDebugMode's privileged
## sidecar); setting it is discouraged and must be temporary.
disableSecurity: false
```

charts/mycarrier-helm/values.schema.json
```diff
@@ -92,6 +92,12 @@
         { "type": "null" }
       ]
     },
+    "disableSecurity": {
```
hpyvovar-mycarrier, line 95: **`disableSecurity` becomes the first root-level property not mirrored in mc-environment — which hard-rejects it downstream** *(technical · important)*

Every other root-level platform property of this chart (`enableVaultCA`, `manualOtelConfig`, `disableOtelAutoinstrumentation`, `tolerations`) appears at three sites in `charts/mc-environment/values.schema.json` — the definitions block (~319-321, remote `$ref` to this file), root properties (~435-437), and the per-environment override object (~496-498) — and is forwarded to the child Application in `templates/appset.yaml:28-32`. `disableSecurity` appears at none of them.

Because `environments[].items` declares `additionalProperties: false` (~line 466), setting `disableSecurity` on an environment entry is a **hard validation failure**, not a silent no-op (verified by render: `at '/environments/0': additional properties 'disableSecurity' not allowed`), and `appset.yaml` builds child values from an explicit allowlist with no passthrough — so the key would never reach mycarrier-helm even if validation passed. An operator on the mc-environment/ApplicationSet delivery path cannot set the flag this PR documents; they would hit a schema rejection mid-incident.

Cheapest fix: state in the docs that `disableSecurity` is settable only on directly-managed mycarrier-helm value files, not through mc-environment `environments[]`. Making a break-glass switch unreachable from committed declarative config may well be the correct design — but then it should be written down as intent rather than left as an accident of omission. Plumbing it through (3 schema sites plus one `disableSecurity: {{ $env.disableSecurity | default false }}` line in appset.yaml) is the alternative only if the flag is meant to be declaratively settable.

charts/mycarrier-helm/values.yaml
pkuzmenko-mycarrier: Good catch on all three points — each was verified against source before rewording, and all three doc sites (values.yaml, schema description, README row) now carry the corrected picture. Fixed in ae8bed2b.

1. **Kyverno backstop**: confirmed. `require-run-as-nonroot` (and `deny-privilege-escalation`) are `validationFailureAction: Enforce` on the `development` branch but `Audit` on `production-csp` and `data`. So admission actually blocks unhardened pods only on dev — exactly where nobody needs break-glass — and the docs now say so, so nobody assumes a prod safety net that isn't there.

2. **Blast radius**: confirmed — both securityContext helpers are included by all six workload templates (deployment, rollout, statefulset, cronjob, job, triggertestengine), so `disableSecurity: true` unhardens every pod in the release and can't be scoped to one application. Now stated explicitly.

3. **"Sole remaining un-harden path"** was too broad, agreed: `applications.<name>.enableDebugMode` injects a `privileged: true` sidecar on the deployment path (`_debug.tpl`, wired at `_spec_deployment.tpl:187`). All three sites now say "the only flag that removes the securityContext hardening block" and point at the debug-sidecar caveat.

Your suggested values.yaml text was applied nearly verbatim, plus one extra line covering the mc-environment restriction from the other thread. One correction to the follow-ups list: a Kyverno capabilities policy does exist — `baseline/disallow_adding_capabilities.yaml` is present on all three branches — though the chart-side `addCapabilities` list is still unbounded, and the `readOnlyRootFilesystem: false` default and the prod/data Enforce graduation are real gaps. All follow-ups (with that correction noted) are filed as bd `mycarrier-c463.35` rather than stretched into this doc-only PR.

charts/mycarrier-helm/values.schema.json
pkuzmenko-mycarrier: Confirmed exactly as described: `environments[].items` declares `additionalProperties: false` and doesn't list `disableSecurity`, and `appset.yaml` forwards an explicit allowlist (`enableVaultCA`, `manualOtelConfig`, `disableOtelAutoinstrumentation`, `tolerations`, `deployment`) with no passthrough — so the key is rejected by schema and would never reach the child chart even if it weren't.

Went with the cheaper option, deliberately: a break-glass switch that can only be set on directly-managed mycarrier-helm value files — not from committed declarative config that Argo would keep re-applying — looks like the right default posture for something this sharp. The docs now state that as intent rather than accident, in all three places (values.yaml comment, schema description, README row). Fixed in ae8bed2b.

If we later decide operators on the ApplicationSet delivery path do need this mid-incident, the plumbing is exactly the three mc-environment schema sites plus one `appset.yaml` forward line you listed — captured in bd `mycarrier-c463.35` so it stays a conscious decision.

hpyvovar-mycarrier (approved):

PR #116, opened 2026-08-17
chore(mycarrier-helm): remove dead lifecycle postStart code

pkuzmenko-mycarrier:
## Summary

Removes dead code from `charts/mycarrier-helm/templates/_lifecycle.tpl` (bd `mycarrier-c463.32`, DEVOPS-176 follow-up to #115):

- **`helm.defaultLifecyclePostStart`** (former lines 7–11) — the `enableVaultCA`-gated `vault.crt` / `update-ca-certificates` postStart hook. Unreachable: its only reference sat inside `helm.lifecycle`, which nothing invoked.
- **`helm.lifecycle`** (former lines 32–47) — every line prefixed `# `. Since `#` is not a comment token in Go text/template, this was actually a *registered* define that nothing called (uncalled, not uncallable) — credit to review for the correction. Either way it contributed nothing to rendered output.

Zero rendered-manifest change, proven by the byte-identical render diff below.

Second commit (review feedback): rewords the prose that described this code as "dead code, fully commented out in `_lifecycle.tpl`" — which would have become a *false*, checkable claim once the code no longer exists. Updated: both `_podsecurity.tpl` hardening-rationale comments, the test title at `tests/security_context_test.yaml:257`, and the `enableVaultCA` description in `values.yaml` + its `README.md` row. Comment/prose text only — no template logic touched.

## Verification

- Chart-wide grep for `helm.lifecycle`, `defaultLifecyclePostStart`, `enableVaultCA`, `vault.crt`, `update-ca-certificates`: **no live consumer**. Remaining `enableVaultCA` references (values.yaml default, mc-environment forwarding, tests) are inert — tests explicitly assert the flag has no effect.
- `helm template` default values, this branch vs a fresh `main` worktree: **byte-identical** (SHA256 match). This is the load-bearing proof — no unittest renders `_lifecycle.tpl`, so test counts alone are coverage-neutral for this change.
- `helm lint charts/mycarrier-helm`: **PASS** (1 chart linted, 0 failed)
- `helm unittest`: mycarrier-helm **638/638 passed** (59 suites, 6/6 snapshots); mc-environment **12/12 passed**

## Notes / out of scope

- `helm.defaultPreStopDelay` and `helm.defaultLifecyclePreStop` remain in place, though now also caller-less within this file — outside dispatched scope; candidate follow-up.
- `enableVaultCA` itself is now consumer-less; removing the flag would touch the values schema and mc-environment forwarding — design/compat decision, candidate follow-up.
- README regeneration via `readme-generator-for-helm` reformats the whole parameter table and pulls in unrelated description drift, so the `enableVaultCA` row was hand-edited to match `values.yaml`; a full regen belongs in its own PR.
- Chart.yaml version bump is bot-automated in this repo ("Automated Change" commits); no manual bump made. Chart publish is a separate step.

**Do not merge without tower approval.**

### DIFF
--- a/charts/mycarrier-helm/templates/_lifecycle.tpl
+++ b/charts/mycarrier-helm/templates/_lifecycle.tpl
@@ -4,12 +4,6 @@
 {{- end }}
 {{- end -}}
 
-{{- define "helm.defaultLifecyclePostStart" -}}
-{{- if .Values.enableVaultCA }}
-{{- printf "curl -k https://vault.infra:8200/v1/pki/ca/pem --output /usr/local/share/ca-certificates/vault.crt;update-ca-certificates" }}
-{{- end }}
-{{- end -}}
-
 {{- define "helm.defaultLifecyclePreStop" -}}
 {{- $language := include "helm.getLanguage" . -}}
 {{- if $language }}
@@ -28,20 +22,3 @@
 {{- printf "pkill vault-env || true" }}
 {{- end }}
 {{- end -}}
-
-# {{- define "helm.lifecycle" -}}
-# {{- $customPostStart := include "helm.defaultLifecyclePostStart" . }}
-# {{- $defaultPreStop := include "helm.defaultLifecyclePreStop" . }}
-# {{- $customPreStopDelay := include "helm.defaultPreStopDelay" . }}
-# {{/* Use dig to safely access nested properties from application context */}}
-# {{- $postStartCommand := dig "lifecycle" "postStart" "echo postStartTest" .application }}
-# {{- $preStopCommand := dig "lifecycle" "preStop" $defaultPreStop .application }}
-#
-# lifecycle:
-#   postStart:
-#     exec:
-#       command: [ "/bin/sh", "-c", {{ printf "%s; %s" $postStartCommand $customPostStart | trim | quote }} ]
-#   preStop:
-#     exec:
-#       command: [ "/bin/sh", "-c", {{ printf "%s%s" $customPreStopDelay $preStopCommand | trim | quote }} ]
-# {{- end -}}
--- a/charts/mycarrier-helm/README.md
+++ b/charts/mycarrier-helm/README.md
@@ -1398,7 +1398,7 @@ When contributing to this chart, please follow the coding standards defined in t
 
 | Name                             | Description                                                                            | Value        |
 | -------------------------------- | -------------------------------------------------------------------------------------- | ------------ |
-| `enableVaultCA`                  | Enable Vault CA certificate download during pod startup (no longer affects securityContext hardening; that postStart hook is currently dead code) | `false`      |
+| `enableVaultCA`                  | Enable Vault CA certificate download during pod startup (no-op: its postStart hook was removed in DEVOPS-176; flag retained for values compatibility) | `false`      |
 | `disableSecurity`                | BREAK-GLASS ONLY (discouraged). When true, every pod in the release renders without runAsNonRoot/securityContext hardening — release-wide, cannot be scoped per-application. Only flag that removes the hardening block (`enableDebugMode`'s privileged sidecar is a separate route). Kyverno `require-run-as-non-root` is Enforce only on development (Audit on production-csp/data). Not settable through mc-environment `environments[]` — direct mycarrier-helm values only. | `false`      |
 | `manualOtelConfig`               | Take over OpenTelemetry configuration from the chart (default false = chart-managed)   | `false`      |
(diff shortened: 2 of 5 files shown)


charts/mycarrier-helm/templates/_lifecycle.tpl
```diff
@@ -27,21 +21,4 @@
 {{- else }}
 {{- printf "pkill vault-env || true" }}
 {{- end }}
-{{- end -}}
-
-# {{- define "helm.lifecycle" -}}
```
hpyvovar-mycarrier, line 32: **Deleting this block makes the `_podsecurity.tpl` hardening rationale positively false, not merely stale** *(technical · important)*

Two live `{{/* */}}` comment blocks in `_podsecurity.tpl` (lines 22-28 above `helm.podSecurityContext`, lines 46-52 above `helm.containerSecurityContext`) record the rationale for a deliberate hardening decision:

> enableVaultCA no longer gates hardening here: its only other consumer, helm.defaultLifecyclePostStart's root-requiring vault.crt postStart hook, is dead code (helm.lifecycle, its sole caller, is fully commented out in _lifecycle.tpl).

After this merge that text makes a positively **false**, directly checkable assertion — `_lifecycle.tpl` will contain no commented-out block at all, and neither named symbol will exist. (The parenthetical was arguably already inaccurate at base: `#` is not a comment token in Go text/template, so `helm.lifecycle` was a registered, callable define — *uncalled*, not *uncallable*. Worth correcting the "fully #-commented / could not execute" wording in the PR body too before squash-merge; also note no unittest exercises `_lifecycle.tpl`, so "638/638" is coverage-neutral evidence — the byte-identical render is the load-bearing proof, which reviewers reproduced independently.)

A maintainer auditing the hardening decision will check the cited file, find the claim false, and have to reconstruct history to decide whether the rationale still holds. Same rot in the test title at `tests/security_context_test.yaml:257` ("...its only other consumer is dead code" — post-merge there is no consumer at all).

Cheap, contained fix (~4 lines of prose): reword both `_podsecurity.tpl` blocks to something like *"enableVaultCA does not gate hardening; it has no template consumer at all (its vault.crt postStart hook was removed in DEVOPS-176). Kyverno require-run-as-non-root is Enforce on dev, so disableSecurity remains the sole explicit break-glass opt-out."* and drop the "is dead code" clause from the test title.

Otherwise this PR verified clean: byte-identical renders reproduced independently (default, `enableVaultCA=true`, and with `applications.*.lifecycle` values set), and removing the last in-repo `curl -k` CA-bootstrap command is a net security improvement.

pkuzmenko-mycarrier: All three points taken — addressed in f7df7d0.

**The false rationale.** You're right, and "stale" was underselling it: once the block is gone, those comments make a claim anyone can check in thirty seconds and find false, which is worse than no comment at all. Both `_podsecurity.tpl` blocks now read "enableVaultCA does not gate hardening here: it has no template consumer at all — its last one, helm.defaultLifecyclePostStart's root-requiring vault.crt postStart hook, was removed as dead code (DEVOPS-176)". Your Kyverno/disableSecurity sentence turned out to already be in the file verbatim, so it stays untouched. The test title at `security_context_test.yaml:257` got the same treatment ("enableVaultCA has no consumer; it does not gate hardening"), and I chased the same rot one step further out: the `enableVaultCA` descriptions in `values.yaml` and its README row also said "currently dead code", so they now say the hook was removed and the flag is retained only for values compatibility.

**`#` is not a comment.** Conceded — Go text/template treats those `#` characters as literal text, so `helm.lifecycle` was a registered, callable define that simply had no caller (and would have emitted `#`-prefixed noise had anyone called it). Uncalled, not uncallable. I've corrected the PR description accordingly.

**Coverage-neutral test counts.** Agreed: no unittest renders `_lifecycle.tpl`, so 638/638 only proves the prose edits broke nothing else. The load-bearing evidence is the render diff, which I re-ran after this commit against a fresh `main` worktree: byte-identical, matching SHA256 on both sides. Thanks for reproducing it independently — that plus your `enableVaultCA=true` and `applications.*.lifecycle` variants is better coverage than the default-values check alone.

One process note: the original dispatch for this task explicitly fenced `_podsecurity.tpl` off, so these edits deliberately stop at comment prose — zero template-logic delta, which the identical render confirms.

One correction to the record: `README.md` regeneration via `readme-generator-for-helm` reformats the entire parameter table and drags in unrelated description drift, so the row was hand-edited to mirror `values.yaml` exactly; a full regen deserves its own PR.

hpyvovar-mycarrier (approved):

PR #118, opened 2026-08-18
feat(mycarrier-helm): fail-closed addCapabilities allow-list guard

pkuzmenko-mycarrier:
## Context
bd `mycarrier-c463.35` part 2 (part 1 was a capability survey — see `standup-notes/2026/08/c463.35-part2-capsurvey.md`). Today `_podsecurity.tpl`'s `addCapabilities` block renders any Linux capability declared under `.application.securityContext.addCapabilities` unconditionally — there is no gate.

## Design (empty default allow-list, fail-closed)
- New values key `.application.securityContext.allowedCapabilities` (list, default `[]`).
- Render-time guard in `charts/mycarrier-helm/templates/_podsecurity.tpl` (the addCapabilities block): for each capability in `addCapabilities`, if it is **not** in `allowedCapabilities`, `fail` with:
  > `Capability "<CAP>" is not permitted by the chart security policy. To permit it, add it to application.securityContext.allowedCapabilities (with security sign-off). Currently allowed: [<list>].`
- `addCapabilities` empty/unset → no-op, renders as today.
- A capability present in both lists → renders as today.
- No default allow-list — a capability is permitted only if the consuming values file explicitly opts in.

## Consumer coupling — result: NONE required
Per the survey, KrakenD was flagged as "the one real consumer" of `addCapabilities` via `NET_BIND_SERVICE`. Verified directly: **KrakenD's rendered manifests (`DevOps/AppCluster-Infrastructure/Config/krakend/deployment.yaml`, `DevOps/KrakenD/deployment.yaml`) are `# Source: krakend/templates/deployment.yaml` — output of a separate `krakend` chart (v0.1.28) with its own hardcoded `securityContext.capabilities.add`.** It does not go through `mycarrier-helm`'s `_podsecurity.tpl` or the `.application.securityContext.addCapabilities` key at all. Same for `charts/hyperdx` (separate `hdx-oss-v2` chart, own `capabilities.add` field).

A repo-wide grep for `addCapabilities` across the whole `mycarrier` working tree found **zero** real (non-test) consumers of this chart's key — only this chart's own `tests/security_context_test.yaml` fixtures used it, and those have been updated in this PR to add `allowedCapabilities` so they keep passing. **No coordinated change is required in any other repo before this chart version is consumed.**

## Schema
`values.schema.json`'s `workloadSecurityContext` definition has `additionalProperties: false`, so `allowedCapabilities` was also added there (alongside `addCapabilities`) — otherwise the new key would be rejected by schema validation.

## Tests added (`charts/mycarrier-helm/tests/security_context_test.yaml`)
- `addCapabilities:[NET_BIND_SERVICE]` without `allowedCapabilities` → render **fails** with the exact guard message (empty allow-list).
- `addCapabilities:[NET_BIND_SERVICE]` with matching `allowedCapabilities` → renders the capability (existing tests updated to add the allow-list).
- `addCapabilities:[SYS_ADMIN]` with a non-empty but non-matching `allowedCapabilities:[NET_BIND_SERVICE]` → **fails**.
- `addCapabilities` unset → no `add` block, no failure (guard is inert).

## Gates
- `helm lint charts/mycarrier-helm`: **PASS** (1 chart linted, 0 failed).
- `helm unittest charts/mycarrier-helm`: **641/641 passed** (59 suites, 6/6 snapshots) — includes the 4 new/updated guard tests.
- `helm template` with default values (no `addCapabilities`), before vs after: **byte-identical** (MD5 match).

## Chart.yaml
No manual version bump — per repo convention (see #115/#116), `chart-version-bumper` CI bumps + publishes automatically on merge to `main`. Publish/propagation to consumers is a separate step from this merge.

### DIFF
--- a/charts/mycarrier-helm/README.md
+++ b/charts/mycarrier-helm/README.md
@@ -1257,7 +1257,8 @@ preserves the chart's existing default rendering — nothing changes until you s
 | Key | Default | Notes |
 | --- | --- | --- |
 | `readOnlyRootFilesystem` | `false` | Sets the container's `readOnlyRootFilesystem`. |
-| `addCapabilities` | `[]` | Applications only. All capabilities are dropped by default (`drop: [ALL]`); list specific capabilities here to add them back. |
+| `addCapabilities` | `[]` | Applications only. All capabilities are dropped by default (`drop: [ALL]`); list specific capabilities here to add them back. Each entry MUST also appear in `allowedCapabilities` (see below) or the render `fail`s. |
+| `allowedCapabilities` | `[]` | Applications only. Chart-security allow-list gating `addCapabilities`. **Empty by default** — no capability is permitted until explicitly allow-listed here (with security sign-off). A capability in `addCapabilities` not present in `allowedCapabilities` fails the render with a clear error message naming the capability and the currently-allowed set. |
 | `fsGroup` | unset | Sets the pod-level `fsGroup`. Must be `>= 1` — `0` is the root group and is rejected by the values schema. **This lower bound is ORG POLICY**, enforced by `MyCarrier-DevOps/kyverno-policies` `require-non-root-groups` (rule `check-fsGroup`: "must be empty or greater than zero") — it is **not** an upstream Pod Security Standards control. The [PSS spec](https://kubernetes.io/docs/concepts/security/pod-security-standards/) contains zero occurrences of `fsGroup`/`runAsGroup`/`supplementalGroups`; those were PodSecurityPolicy fields, and PSP is retired. |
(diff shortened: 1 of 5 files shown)


pkuzmenko-mycarrier:
Closing unmerged per decision: leaving mycarrier-helm as-is for now. Org-wide gh code search found ZERO consumers of this chart's .application.securityContext.addCapabilities, and kyverno disallow_adding_capabilities already denies ALL added caps on all 3 branches (Enforce on dev, Audit on prod-csp/data). A chart-side guard can be added later (hardcode-empty, aligned) or handled via a kyverno exception if a real consumer emerges. Tracked in bd c463.35 part 2.

